PR #3, opened 2016-06-12
docker-compose v2 format example

mosen:
Add example docker-compose file using v2 format.
It isn't exactly perfect especially the volume configuration, but it's a fair reference to all of the ENV var parameters.

### DIFF
--- a/docker-compose-v2.yml
+++ b/docker-compose-v2.yml
@@ -1,0 +1,52 @@
+version: '2'
+services:
+  redis:
+    image: redis
+    restart: always
+    volumes:
+     - redis:/data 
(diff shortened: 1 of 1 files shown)


groob:
Thanks, looks good. 

Technically you don't need `MICROMDM_REDIS_CONN_URL: 'redis:6379'` if you're connecting redis with a link, but adding it is good for documentation. 

```
func getRedisConnFromENV(host string) string {
	port := os.Getenv("REDIS_PORT_6379_TCP_PORT")
	conn := fmt.Sprintf("%v:%v", host, port)
	return conn
}
```

PR #4, opened 2016-06-14
Feature/cors

mosen:
Heya added CORS middleware to the mux if flag `--cors-origin <host>` or env var `MICROMDM_CORS_ORIGIN` is set. Please review, it's honestly the first piece of go code I've put down.

m


groob:
I merged the request in the CLI and updated the vendor packages, but it looks like I should've merged the branch first.

groob:
https://github.com/isaacs/github/issues/2 would be great

PR #7, opened 2016-06-22
Schema and Devices changes

mosen:
Lots of changes for persisting information from DeviceInformation queries.


connect/service.go
```diff
@@ -5,12 +5,13 @@ import (
...
 	"github.com/micromdm/micromdm/device"
 	"github.com/pkg/errors"
+	"golang.org/x/net/context"
 )
 
 // Service defines methods for an MDM service
 type Service interface {
-	Acknowledge(deviceUDID, commandUUID string) (int, error)
-	NextCommand(deviceUDID string) ([]byte, int, error)
+	Acknowledge(ctx context.Context, req mdmConnectRequest) (int, error)
```
groob: The `req` argument should be the exported mdm.Response instead of the command request. Having an un-exported type here makes it hard to call Service outside the package.

connect/service.go
```diff
@@ -63,3 +64,8 @@ func (svc service) checkRequeue(deviceUDID string) (int, error) {
 	}
 	return 0, nil
 }
+
+// Acknowledge Queries sent with DeviceInformation command
+func (svc service) ackQueryResponses() (error) {
```
groob: Not sure I understand what this is for.

Also, `(error)` should just be `error`. By convention you only need to surround it with `()` if there is more than one arg.

connect/service.go
```diff
@@ -26,13 +27,31 @@ type service struct {
...
 }
 
-func (svc service) Acknowledge(deviceUDID, commandUUID string) (int, error) {
-	total, err := svc.commands.DeleteCommand(deviceUDID, commandUUID)
+func (svc service) Acknowledge(ctx context.Context, req mdmConnectRequest) (int, error) {
+	var err error
+
+	switch req.RequestType {
+	case "DeviceInformation":
+		err = svc.ackQueryResponses(req)
```
groob: Instead of declaring `var err error` at the top, I would instead write this as 
```
if err := svc.ackQueryResponses(req); err !=nil {
    return 0, err
}
```

connect/middleware.go
```diff
@@ -0,0 +1,28 @@
...
+
+func NewMiddleware(datastore device.Datastore, next Service) (*lastCheckinMiddleware) {
+	return &lastCheckinMiddleware{
+		datastore,
+		next,
+	}
+}
+
+func (mw lastCheckinMiddleware) Acknowledge(ctx context.Context, req mdmConnectRequest) (int, error) {
+	mw.devices.UpdateDeviceCheckinByUDID(req.UDID)
```
groob: This should not ignore the error return.

connect/middleware.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/micromdm/micromdm/device"
+	"golang.org/x/net/context"
+)
+
+type lastCheckinMiddleware struct {
+	devices device.Datastore
+	next    Service
+}
+
+func NewMiddleware(datastore device.Datastore, next Service) (*lastCheckinMiddleware) {
```
groob: Your `Acknowledge` function already uses `ackQueryResponses()` which I think is the correct place to persist the data, instead of wrapping it in middleware.  

Is this a holdover from an earlier commit?

connect/service.go
groob: Nevermind, I was on an earlier diff.

groob:
I made some comments on the PR, would you mind making these changes?

* All business logic should happen within the main Service. I'd reserve wrapping calls in a middleware for things like logging and instrumentation. 
* The service methods should take exported types as arguments. This means changing Acknowledge to take the mdm.Response struct rather than the private `mdmConnectRequest` struct. 

Otherwise I like the commit. Good work!

groob:
Another thought - this one I'm not 100% sure about myself so feel free to ignore it. 
```
	UpdateDeviceQueryResponseByUDID(udid string, responses mdm.QueryResponses) error
	UpdateDeviceCheckinByUDID(udid string) error
```
Both of these can be accomplished using the `Save(msg string, dev *Device) error` method on the Datastore. You'd only need to add a `QueryResponseData []byte` field to the Device struct. 

My preference for using the Save method is that it would avoid adding a whole new method for each database operation, and there are a lot of other MDM responses that we're likely to want to persist in the future. 

This is fine for now if you don't want to change it.

PR #8, opened 2016-06-29
Mdm enroll

mosen:
Enrollment profile via /mdm/enroll.

I think this actually includes some commits from mosen/micromdm develop branch since I rebased mdm_enroll on to develop.

### DIFF
--- a/device/device.go
+++ b/device/device.go
@@ -71,7 +71,7 @@ func (status *DEPProfileStatus) Scan(value interface{}) error {
 		return nil
 	}
 	if sv, err := driver.String.ConvertValue(value); err == nil {
-		if v, ok := sv.([]byte); ok {
+		if v, ok := sv.(string); ok {
 			*status = DEPProfileStatus(v)
 			return nil
 		}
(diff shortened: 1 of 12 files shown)


device/device.go
```diff
@@ -6,7 +6,7 @@ import (
 	"time"
 
 	"github.com/micromdm/dep"
-	"github.com/micromdm/mdm"
+	//"github.com/micromdm/mdm"
```
groob: Remove commented out import path.

enroll/service.go
```diff
@@ -0,0 +1,100 @@
...
+
+	scepSubject := [][][]string{
+		[][]string{
+			[]string{"O", "MicroMDM"},
+			[]string{"CN", "MDM Identity Certificate:UDID"},
+		},
+	}
+
+	return &service{
+		Url:         "https://micromdm.local:6443",
```
groob: by convention, initialisms should be all capitals https://github.com/golang/go/wiki/CodeReviewComments#initialisms

groob: The URL for the service and SCEP server should be configurable using a flag in main.

groob:
merged to master. I rebased, which usually closes the PR, but somehow not in this case. 

Thanks again for all the work. 👍

PR #10, opened 2016-07-12
enrollment endpoint, applications table/datastore and command failure handling

mosen:
Recent bits and pieces including the code to register handlers for Acknowledge requests.


certificates/certificate.go
```diff
@@ -1,13 +1,11 @@
 package certificates
 
-import (
-	"crypto/x509"
-)
+import ()
```
groob: Is this something InteliJ does? I believe goimports itself would just remove the line if the are no imports.

mosen:
I am in the process of cherry picking commits into different "topics" of functionality.
Please leave this open just so I can have a quick guide to the commit refs.
PR will be closed once all items have been cherry picked.

PR #13, opened 2016-10-06
Enrollment

mosen:
A compilation of cherry picked commits regarding the enrollment endpoint and enrollment configuration variables.

Some changes to CORS, installed applications, and installed certificates were edited out.

Mostly this builds on the original enroll endpoint which had some hard coded values and makes them all configurable.

### DIFF
--- a/enroll/endpoint.go
+++ b/enroll/endpoint.go
@@ -5,19 +5,25 @@ import (
 	"golang.org/x/net/context"
 )
 
+type Endpoints struct {
+	GetEnrollEndpoint endpoint.Endpoint
+}
+
 type mdmEnrollRequest struct{}
 
 type mdmEnrollResponse struct {
 	Profile
 }
 
-func makeEnrollEndpoint(svc Service) endpoint.Endpoint {
+func MakeServerEndpoints(s Service) Endpoints {
+	return Endpoints{
+		GetEnrollEndpoint: MakeGetEnrollEndpoint(s),
+	}
+}
+
+func MakeGetEnrollEndpoint(s Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
-		//req := request.(mdmEnrollRequest)
-		profile, err := svc.Enroll()
-		if err != nil {
-			return mdmEnrollResponse{}, err
-		}
-		return mdmEnrollResponse{profile}, nil
+		profile, err := s.Enroll(ctx)
+		return mdmEnrollResponse{profile}, err
 	}
 }
--- a/enroll/endpoint.go
+++ b/enroll/endpoint.go
@@ -12,7 +12,8 @@ type Endpoints struct {
 type mdmEnrollRequest struct{}
 
 type mdmEnrollResponse struct {
-	Profile
+	Profile,
+	Err error `plist:"error,omitempty"`
 }
 
 func MakeServerEndpoints(s Service) Endpoints {
@@ -24,6 +25,6 @@ func MakeServerEndpoints(s Service) Endpoints {
 func MakeGetEnrollEndpoint(s Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		profile, err := s.Enroll(ctx)
-		return mdmEnrollResponse{profile}, err
+		return mdmEnrollResponse{profile, err}, nil
 	}
 }
--- a/enroll/service.go
+++ b/enroll/service.go
@@ -90,6 +90,7 @@ func (svc service) Enroll() (Profile, error) {
 		caPayload := NewPayload("com.apple.ssl.certificate")
 		caPayload.PayloadDisplayName = "Root certificate for MicroMDM"
 		caPayload.PayloadDescription = "Installs the root CA certificate for MicroMDM"
+		caPayload.PayloadIdentifier = "com.github.micromdm.ssl.ca"
 		caPayload.PayloadContent = svc.CACert
 
 		profile.PayloadContent = []interface{}{*scepPayload, mdmPayloadContent, *caPayload}
--- a/enroll/service.go
+++ b/enroll/service.go
@@ -6,7 +6,7 @@ type Service interface {
 	Enroll() (Profile, error)
 }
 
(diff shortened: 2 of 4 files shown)


enroll/service.go
```diff
@@ -98,3 +103,58 @@ func (svc service) Enroll() (Profile, error) {
 
 	return *profile, nil
 }
+
+type service struct {
+	Url           string
```
groob: `Url` should be `URL`. This applies to all initialisms like `Id` vs` ID` or `Smtp` vs `SMTP` etc.

enroll/service.go
```diff
@@ -98,3 +103,58 @@ func (svc service) Enroll() (Profile, error) {
...
+	SCEPUrl       string
+	SCEPChallenge string
+	SCEPSubject   [][][]string
+	Topic         string // APNS Topic for MDM notifications
+}
+
+func (svc service) Enroll() (Profile, error) {
```
groob: I'm confused. a bit earlier you added a `Enroll` method with a `ctx` argument. This one is different?

main.go
```diff
@@ -39,27 +39,28 @@ func main() {
...
-		flEnrollment   = flag.String("profile", envString("MICROMDM_ENROLL_PROFILE", ""), "path to enrollment profile")
-		flDEPCK        = flag.String("dep-consumer-key", envString("DEP_CONSUMER_KEY", ""), "dep consumer key")
-		flDEPCS        = flag.String("dep-consumer-secret", envString("DEP_CONSUMER_SECRET", ""), "dep consumer secret")
-		flDEPAT        = flag.String("dep-access-token", envString("DEP_ACCESS_TOKEN", ""), "dep access token")
-		flDEPAS        = flag.String("dep-access-secret", envString("DEP_ACCESS_SECRET", ""), "dep access secret")
-		flDEPsim       = flag.Bool("depsim", envBool("DEP_USE_DEPSIM"), "use default depsim credentials")
-		flDEPServerURL = flag.String("dep-server-url", envString("DEP_SERVER_URL", ""), "dep server url. for testing. Use blank if not running against depsim")
-		flPkgRepo      = flag.String("pkg-repo", envString("MICROMDM_PKG_REPO", ""), "path to pkg repo")
-		flCorsOrigin   = flag.String("cors-origin", envString("MICROMDM_CORS_ORIGIN", ""), "allowed domain for cross origin resource sharing")
+		flUrl           = flag.String("url", envString("MICROMDM_URL", ""), "public facing url")
```
groob: `flURL` not `flUrl`

main.go
```diff
@@ -39,27 +39,28 @@ func main() {
...
+		flPushPass      = flag.String("push-pass", envString("MICROMDM_PUSH_PASS", ""), "push certificate password")
+		flEnrollment    = flag.String("profile", envString("MICROMDM_ENROLL_PROFILE", ""), "path to enrollment profile")
+		flDEPCK         = flag.String("dep-consumer-key", envString("DEP_CONSUMER_KEY", ""), "dep consumer key")
+		flDEPCS         = flag.String("dep-consumer-secret", envString("DEP_CONSUMER_SECRET", ""), "dep consumer secret")
+		flDEPAT         = flag.String("dep-access-token", envString("DEP_ACCESS_TOKEN", ""), "dep access token")
+		flDEPAS         = flag.String("dep-access-secret", envString("DEP_ACCESS_SECRET", ""), "dep access secret")
+		flDEPsim        = flag.Bool("depsim", envBool("DEP_USE_DEPSIM"), "use default depsim credentials")
+		flDEPServerURL  = flag.String("dep-server-url", envString("DEP_SERVER_URL", ""), "dep server url. for testing. Use blank if not running against depsim")
+		flPkgRepo       = flag.String("pkg-repo", envString("MICROMDM_PKG_REPO", ""), "path to pkg repo")
+		flCorsOrigin    = flag.String("cors-origin", envString("MICROMDM_CORS_ORIGIN", ""), "allowed domain for cross origin resource sharing")
```
groob: should be `flCORSOrigin`

main.go
```diff
@@ -207,10 +208,17 @@ func main() {
 	mux.Handle("/mdm/connect", connectHandler)
 
 	if checkEmptyArgs(*flUrl, *flScepUrl) {
-		logger.Log("warn", "Enrollment endpoint /mdm/enroll will be disabled because you did not specify flags for the external URL or SCEP URL")
+		logger.Log("warn", "Enrollment endpoint /mdm/enroll will be disabled because you did not specify flags/environment vars for the external URL (--url MICROMDM_URL) or SCEP URL (--scep-url/MICROMDM_SCEP_URL)")
```
groob: Is one of these options required? If so ask the user to enter the appropriate flag and exit.

enroll/endpoint.go
```diff
@@ -5,19 +5,25 @@ import (
...
+func MakeGetEnrollEndpoint(s Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
-		//req := request.(mdmEnrollRequest)
-		profile, err := svc.Enroll()
-		if err != nil {
-			return mdmEnrollResponse{}, err
-		}
-		return mdmEnrollResponse{profile}, nil
+		profile, err := s.Enroll(ctx)
+		return mdmEnrollResponse{profile}, err
```
groob: The correct way to handle the error would be to add it to the response struct. This is because it's a business logic error(as opposed to a http transport error), and the server will encode the error response to the client. 
Add
```
Err     error `plist:"error,omitempty"
```
to the `mdmEnrollResponse` struct. 

I'm ok with merging this and fixing in a separate PR if you open an issue so we don't forget.

enroll/service.go
mosen: I realise this now after having read more about golang :)

main.go
mosen: This was implemented in such a way that you could enroll with SCEP or use the same client identity for every single device (like in a test scenario).

enroll/endpoint.go
mosen: Yeah my bad, I'm not totally understanding the flow of error handling in gokit/http.

PR #14, opened 2016-10-06
Installed certificates

mosen:
This PR includes most of the functionality related to processing installed certificates responses.
Note that you cannot easily determine which certificates have changed from the previous response, so the entire certificate inventory per-device is replaced when a new response arrives.

There are some kludgy hacks because parts of the applications service made it into some of these cherry picks and will have to be edited out if you want a "pure" feature branch.

### DIFF
--- a/certificates/certificate.go
+++ b/certificates/certificate.go
@@ -1,0 +1,12 @@
+package certificates
+
+import (
+	"crypto/x509"
+)
+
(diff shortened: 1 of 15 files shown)


mosen:
This branch PR has several conflicts so I will be creating a new PR for the changes included in this branch.

PR #16, opened 2016-10-06
Commands management

mosen:
These commits represent the addition of datastore method + endpoint for listing pending commands.

It also replaces the response handling switch block by matching the command UUID of the response with an outstanding command UUID instead of switching on RequestType. This is more reliable because iOS devices do not reply with the RequestType of the request that generated the response.

Also kludge

### DIFF
--- a/command/datastore.go
+++ b/command/datastore.go
@@ -28,6 +28,7 @@ type Datastore interface {
 	QueueCommand(deviceUDID, commandUUID string) error
 	NextCommand(deviceUDID string) ([]byte, int, error)
 	DeleteCommand(deviceUDID, commandUUID string) (int, error)
+	Commands(deviceUDID string) ([]mdm.Payload, error)
 }
 
 //NewDB creates a Datastore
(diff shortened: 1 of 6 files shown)


command/datastore_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"testing"
+)
+
+var (
+	ds     Datastore
+	logger log.Logger
+	err    error
+)
+
+//func setup() {
```
groob: Delete instead of commenting lines.

command/datastore_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"github.com/go-kit/kit/log"
+	"github.com/micromdm/mdm"
+	"os"
+	"testing"
+)
+
+var (
```
groob: Let's avoid having global vars here. Create a setup function to get a datastore

command/transport.go
```diff
@@ -81,6 +89,18 @@ func decodeDeleteCommandRequest(_ context.Context, r *http.Request) (interface{}
...
 }
 
+func decodeGetCommandsRequest(_ context.Context, r *http.Request) (interface{}, error) {
+	vars := mux.Vars(r)
+	udid, ok := vars["udid"]
+	if !ok {
+		return nil, errBadRouting
+	}
+
+	var request getCommandsRequest
```
groob: I would prefer if this was `getCommandsRequest{UDID: udid}` instead

command/datastore.go
```diff
@@ -124,6 +126,48 @@ func (rds redisDB) DeleteCommand(deviceUDID, commandUUID string) (int, error) {
 	return total, nil
 }
 
+func (rds redisDB) Commands(deviceUDID string) ([]mdm.Payload, error) {
+	conn := rds.pool.Get()
+	defer conn.Close()
+
+	commandUuids, err := redis.Values(conn.Do("LRANGE", deviceUDID, "0", "-1"))
```
groob: Uppercase `Uuid` initialism here.

connect/service.go
```diff
@@ -3,6 +3,8 @@ package connect
 import (
 	"encoding/json"
 	"github.com/micromdm/mdm"
+	apps "github.com/micromdm/micromdm/applications"
```
groob: Go package names should always be singular. I'll create an issue once this is merged to update.

mosen: Will rename applications after merge.

mosen:
There are merge conflicts so these will be resolved in another PR.
All of the fixes have been applied

PR #18, opened 2016-10-06
Installed application list

mosen:
These commits relate to the installed applications service for acknowledging and querying lists of installed applications on a per device basis.

There's no unique constraint that will work for applications because some bundles have a name and no version etc. Therefore the entire list is replaced every time the device updates it.

Also more kludge/bits and pieces that should be in other branches.

### DIFF
--- a/applications/datastore.go
+++ b/applications/datastore.go
@@ -6,7 +6,6 @@ import (
 	"github.com/jmoiron/sqlx"
 	_ "github.com/lib/pq" // postgres driver
 	"github.com/pkg/errors"
-	"strings"
 	"time"
 )
 
@@ -14,7 +13,7 @@ import (
 type Datastore interface {
 	New(a *Application) (string, error)
 	Applications(params ...interface{}) ([]Application, error)
-	GetApplicationsByDeviceUUID(deviceUUID string) (*[]Application, error)
+	GetApplicationsByDeviceUUID(deviceUUID string) ([]Application, error)
 	SaveApplicationByDeviceUUID(deviceUUID string, app *Application) error
 }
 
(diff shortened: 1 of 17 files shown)


applications/datastore_test.go
```diff
@@ -0,0 +1,254 @@
...
+		Version:      sql.NullString{"2.9.1", true},
+	},
+	{ // macOS Application with no bundle identifier
+		UUID:       "84c78174-8331-4cf3-98c3-a4b1434617e5",
+		Name:       "Wireless Network Utility",
+		BundleSize: sql.NullInt64{2416111, true},
+	},
+}
+
+var logger log.Logger = log.NewNopLogger()
```
groob: Remove the global.

applications/datastore_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+//func TestNewDB(t *testing.T) {
```
groob: Remove commented code.

connect/service.go
```diff
@@ -120,3 +148,53 @@ func (svc service) ackQueryResponses(req mdm.Response) error {
...
+		}
+
+		if err := svc.apps.NewDeviceApp(&newApp); err != nil {
+			fmt.Println(err)
+			fmt.Printf("%v\n", newApp)
+			return fmt.Errorf("inserting an application for device: %s", err)
+		}
+
+		requestApps[i] = newApp
+		//fmt.Printf("%v\n", newApp)
```
groob: remove commented code.

connect/service_test.go
```diff
@@ -0,0 +1,155 @@
...
+	return &mdm.Payload{}, nil
+}
+func (mc MockCmd) NextCommand(udid string) ([]byte, int, error) {
+	return []byte{}, 0, nil
+}
+func (mc MockCmd) DeleteCommand(deviceUDID, commandUUID string) (int, error) {
+	return 0, nil
+}
+
+var ctx context.Context
```
groob: Set these up without using globals.

applications/datastore.go
```diff
@@ -0,0 +1,166 @@
...
+	SaveApplicationByDeviceUUID(deviceUUID string, app *Application) error
+	DeleteDeviceApplications(deviceUUID string) error
+}
+
+type pgStore struct {
+	*sqlx.DB
+	logger kitlog.Logger
+}
+
+func NewDatastore(connection *sqlx.DB, logger kitlog.Logger) (Datastore, error) {
```
groob: What's the difference between `NewDB` and `NewDatastore`?
Why are both needed? 

At the very least add a comment on top of each to add an explanation.

PR #29, opened 2016-10-12
Feature/certificate expiry warnings

mosen:
Added a few warnings to stop myself from using old or wrong certificates. Figured it would be good to catch this for other users as well. Might also add hostname verification at a later date.


main.go
```diff
@@ -333,11 +335,35 @@ func checkEmptyArgs(args ...string) bool {
...
+			os.Exit(1)
+		}
+
+		cert, err := x509.ParseCertificate(chain.Certificate[0]) // Leaf is always the first entry
+		if err != nil {
+			logger.Log("err", "error parsing TLS certificate")
+			os.Exit(1)
+		}
+
+		if _, err := cert.Verify(x509.VerifyOptions{}); err != nil {
```
groob: The documentation for `Verify` says 
> If opts.Roots is nil and system roots are unavailable the returned error will be of type SystemRootsError.

Are you sure you're not going to run into the error by using an empty `VerifyOptions` struct?

Otherwise looks good.

main.go
```diff
@@ -333,11 +335,35 @@ func checkEmptyArgs(args ...string) bool {
...
+			switch e := err.(type) {
+			case x509.CertificateInvalidError:
+				switch e.Reason {
+				case x509.Expired:
+					logger.Log("err", "certificate has expired")
```
groob: We want to exit after an error, no?

main.go
mosen: I'm not sure but this is lifted from buford verbatim :)
I don't know what extra roots i could provide other than our included CA?

main.go
mosen: MicroMDM now exits if certificate validation fails for any reason.
Host name validation is not yet done.

PR #31, opened 2016-10-12
Hotfix/enroll payloadscope

mosen:
The `PayloadScope` attribute of the profile struct was not properly being serialized which resulted in enrollments being User scoped only. This fixes the serialization of that attribute and a few others.

Verified new profiles are enrolled in the system scope.


mosen:
I have no idea why the changes from the previous PR are showing in this revision but im merging anyway.

PR #44, opened 2016-11-06
Feature/dep optional

mosen:
depClient() function to construct a client may now return an error as second argument, indicating that the DEP client is not available.
management.Service will return an error "feature not available" if you attempt to use dep fetch and don't supply a DEP client.
management test teardown via gomigrate rollback
new management test ensures that constructing the management service without a dep client actually works and the endpoint does not cause a panic.

### DIFF
--- a/main.go
+++ b/main.go
@@ -57,6 +57,7 @@ func main() {
 		flPushCert      = flag.String("push-cert", envString("MICROMDM_PUSH_CERT", ""), "path to push certificate")
 		flPushPass      = flag.String("push-pass", envString("MICROMDM_PUSH_PASS", ""), "push certificate password")
 		flEnrollment    = flag.String("profile", envString("MICROMDM_ENROLL_PROFILE", ""), "path to enrollment profile")
+		flDEP           = flag.Bool("dep", envBool("MICROMDM_USE_DEP"), "use DEP")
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -211,8 +211,15 @@ func main() {
 		os.Exit(1)
 	}
 
-	dc := depClient(logger, *flDEPCK, *flDEPCS, *flDEPAT, *flDEPAS, *flDEPServerURL, *flDEPsim)
-	mgmtSvc := management.NewService(deviceDB, workflowDB, dc, pushSvc, appsDB, certsDB)
+	dc, err := depClient(logger, *flDEPCK, *flDEPCS, *flDEPAT, *flDEPAS, *flDEPServerURL, *flDEPsim)
+	var mgmtSvc management.Service
+	if err != nil {
```
groob: You're suppressing the error here, there's potential for the client to fail for other reasons besides invalid credentials.
If we want `dep` to be opt-in, I'd rather start without depClient by default. Have the user pass a flag with DEP credentials, otherwise start the server without DEP.

mosen: would you be happy with a --dep boolean flag?

mosen:
Added --dep boolean flag. Default status is disabled.
Console will output a message letting the user know whether the DEP feature is disabled or enabled.

PR #47, opened 2016-11-11
Feature/dep sync

mosen:
Initial draft DEP Sync

### DIFF
--- a/depsync/syncer.go
+++ b/depsync/syncer.go
@@ -1,4 +1,4 @@
-package dep
+package depsync
 
 import (
 	"errors"
@@ -31,13 +31,15 @@ type syncer struct {
 	doneChan             <-chan struct{}
 	Cursor               *Cursor
 	InitialFetchComplete bool
+	errorCount           int
 }
 
-func NewSyncer(client dep.Client, logger log.Logger, tickerChan <-chan time.Time, done <-chan struct{}) Syncer {
+func NewSyncer(client dep.Client, logger log.Logger, tickerChan <-chan time.Time) Syncer {
 	return &syncer{
 		logger:     logger,
 		client:     client,
 		tickerChan: tickerChan,
+		errorCount: 0,
 	}
 }
 
@@ -107,11 +109,14 @@ func (s *syncer) Sync(deviceChan chan dep.Device) (bool, error) {
 // The first run will always download a complete list of devices. Deltas will be fetched if the process is still running
 // after the first sync interval
 func (s *syncer) Start(deviceChan chan dep.Device) {
+	s.logger.Log("level", "debug", "msg", "DEP sync routine started")
+
 	for range s.tickerChan {
 		if !s.InitialFetchComplete {
 			more, err := s.Fetch(deviceChan)
 			if err != nil {
 				s.logger.Log("level", "error", "msg", fmt.Sprintf("Fetching initial snapshot of devices from DEP: %v", err))
+				s.errorCount++
 			} else {
 				s.logger.Log("level", "debug", "msg", fmt.Sprintf("More devices after this batch: %t", more))
 				if !more {
--- a/depsync/syncer.go
+++ b/depsync/syncer.go
@@ -3,50 +3,49 @@ package depsync
 import (
 	"errors"
 	"fmt"
+	"time"
+
 	"github.com/go-kit/kit/log"
 	"github.com/micromdm/dep"
-	"time"
 )
 
 // Syncer is responsible for fetching devices from DEP, keeping them in sync, and storing cursor information about the
 // current fetch operation.
 type Syncer interface {
-	Start(deviceChan chan dep.Device)
-	Pause()
-	Resume()
-	Fetch(deviceChan chan dep.Device) (bool, error)
+	Start(deviceChan chan<- dep.Device)
+	Fetch(deviceChan chan<- dep.Device) (bool, error)
 }
 
 // Cursor represents information about the current fetch operation. It is required to fetch multiple "pages" of devices.
 type Cursor struct {
 	Value   string
-	Created *time.Time // >7 days cursor is invalid.
+	Created time.Time // >7 days cursor is invalid.
 }
 
 type syncer struct {
 	logger               log.Logger
 	client               dep.Client
-	tickerChan           <-chan time.Time
-	tickerChanDisabled   <-chan time.Time
-	doneChan             <-chan struct{}
+	fetchInterval        time.Duration
+	ticker               *time.Ticker
+	done                 <-chan struct{}
 	Cursor               *Cursor
 	InitialFetchComplete bool
 	errorCount           int
 }
 
-func NewSyncer(client dep.Client, logger log.Logger, tickerChan <-chan time.Time) Syncer {
+func NewSyncer(client dep.Client, logger log.Logger, fetchInterval time.Duration, done <-chan struct{}) Syncer {
 	return &syncer{
-		logger:     logger,
-		client:     client,
-		tickerChan: tickerChan,
-		errorCount: 0,
+		logger:        logger,
+		client:        client,
+		fetchInterval: fetchInterval,
+		errorCount:    0,
+		done:          done,
 	}
 }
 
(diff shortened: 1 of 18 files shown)


depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+	"fmt"
+	"github.com/go-kit/kit/log"
+	"github.com/micromdm/dep"
+	"time"
+)
+
+// Syncer is responsible for fetching devices from DEP, keeping them in sync, and storing cursor information about the
+// current fetch operation.
+type Syncer interface {
+	Start(deviceChan chan dep.Device)
```
groob: no neet do name things `fooChan`. The channel already says what it is.

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+	Pause()
+	Resume()
+	Fetch(deviceChan chan dep.Device) (bool, error)
+}
+
+// Cursor represents information about the current fetch operation. It is required to fetch multiple "pages" of devices.
+type Cursor struct {
+	Value   string
+	Created *time.Time // >7 days cursor is invalid.
```
groob: Should just be `time.Time`, no?

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+type syncer struct {
+	logger               log.Logger
+	client               dep.Client
+	tickerChan           <-chan time.Time
```
groob: lets add `fetchInterval time.Duration` and create an internal ticker instead. Both this and disabled seem unnecessary.

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+		logger:     logger,
+		client:     client,
+		tickerChan: tickerChan,
+		errorCount: 0,
+	}
+}
+
+// Fetch fetches a list of DEP devices in batches. The first return value indicates that there are more devices to fetch.
+func (s *syncer) Fetch(deviceChan chan dep.Device) (bool, error) {
+	var deviceResponse *dep.DeviceResponse
```
groob: group the vars with `var ( )`

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+	var err error
+
+	if s.Cursor != nil {
```
groob: instead of using `else`, how about we `append` to an opts slice and then use `cilent.FetchDevices(opts...)`

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+func NewSyncer(client dep.Client, logger log.Logger, tickerChan <-chan time.Time) Syncer {
+	return &syncer{
+		logger:     logger,
+		client:     client,
+		tickerChan: tickerChan,
+		errorCount: 0,
+	}
+}
+
+// Fetch fetches a list of DEP devices in batches. The first return value indicates that there are more devices to fetch.
```
groob: Wrap the doc string at 80 columns.

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+	Value   string
+	Created *time.Time // >7 days cursor is invalid.
+}
+
+type syncer struct {
+	logger               log.Logger
+	client               dep.Client
+	tickerChan           <-chan time.Time
+	tickerChanDisabled   <-chan time.Time
+	doneChan             <-chan struct{}
```
groob: `done chan struct{}` ? `doneChan` stutters.

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+	return deviceResponse.MoreToFollow, nil
+}
+
+// Start starts the synchronisation schedule which runs at a configured interval.
+// The first run will always download a complete list of devices. Deltas will be fetched if the process is still running
+// after the first sync interval
+func (s *syncer) Start(deviceChan chan dep.Device) {
+	s.logger.Log("level", "debug", "msg", "DEP sync routine started")
+
+	for range s.tickerChan {
```
groob: you can just have the channel block until it receives a tick `<- ticker`

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+	return &syncer{
+		logger:     logger,
+		client:     client,
+		tickerChan: tickerChan,
+		errorCount: 0,
+	}
+}
+
+// Fetch fetches a list of DEP devices in batches. The first return value indicates that there are more devices to fetch.
+func (s *syncer) Fetch(deviceChan chan dep.Device) (bool, error) {
```
groob: `deviceChan` should be renamed to `device` and the channel input needs a direction specified. Is this a send or receive channel?

depsync/writer.go
```diff
@@ -0,0 +1,70 @@
...
+
+func NewWriter(datastore device.Datastore, logger log.Logger) Writer {
+	return &writer{
+		datastore: datastore,
+		logger:    logger,
+	}
+}
+
+func (w *writer) Start(deviceChan <-chan dep.Device) {
+	for dev := range deviceChan {
```
groob: This looks like a candidate for `select`

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+			} else {
+				s.logger.Log("level", "debug", "msg", fmt.Sprintf("More devices after this sync: %t", more))
+			}
+		}
+	}
+}
+
+// Pause pauses the synchronisation timer
+func (s *syncer) Pause() {
+	s.tickerChanDisabled = s.tickerChan
```
groob: This looks race-y. try to compile with `go build -race` and see if it panics after running for a bit. 

IMO it's also not necessary if we replace the channel with just a time.Duration interval.

main.go
```diff
@@ -43,30 +45,31 @@ func main() {
...
+		flPushPass        = flag.String("push-pass", envString("MICROMDM_PUSH_PASS", ""), "push certificate password")
+		flEnrollment      = flag.String("profile", envString("MICROMDM_ENROLL_PROFILE", ""), "path to enrollment profile")
+		flDEP             = flag.Bool("dep", envBool("MICROMDM_USE_DEP"), "use DEP")
+		flDEPCK           = flag.String("dep-consumer-key", envString("DEP_CONSUMER_KEY", ""), "dep consumer key")
+		flDEPCS           = flag.String("dep-consumer-secret", envString("DEP_CONSUMER_SECRET", ""), "dep consumer secret")
+		flDEPAT           = flag.String("dep-access-token", envString("DEP_ACCESS_TOKEN", ""), "dep access token")
+		flDEPAS           = flag.String("dep-access-secret", envString("DEP_ACCESS_SECRET", ""), "dep access secret")
+		flDEPsim          = flag.Bool("depsim", envBool("DEP_USE_DEPSIM"), "use default depsim credentials")
+		flDEPServerURL    = flag.String("dep-server-url", envString("DEP_SERVER_URL", ""), "dep server url. for testing. Use blank if not running against depsim")
+		flDEPSyncInterval = flag.String("dep-sync-interval", envString("DEP_SYNC_INTERVAL", "60"), "dep synchronization interval in seconds.")
```
groob: use `flag.Duration` and then just pass that down :)

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
+package depsync
+
+import (
```
groob: Not sure what IDE you're using, but the imports need to be sorted - stdlib imports first, then a space, then other imports. `goimports` will do that for you.

depsync/syncer.go
```diff
@@ -0,0 +1,149 @@
...
+	s.logger.Log("level", "debug", "msg", fmt.Sprintf("Fetching %d devices", len(deviceResponse.Devices)))
+	for _, dev := range deviceResponse.Devices {
+		deviceChan <- dev
+	}
+
+	return deviceResponse.MoreToFollow, nil
+}
+
+// Sync fetches devices since the given cursor value
+func (s *syncer) Sync(deviceChan chan dep.Device) (bool, error) {
```
groob: I have no clue what `bool` is. Consider naming the return values here.

Issue #55, opened 2016-11-14
app: allow passing separate APNS certificate and private key. 


groob:
Currently the flags menu reads:
```
  -push-cert string
        path to push certificate
  -push-key string
        path to push certificate private key(if not using a single .p12 file)
  -push-password string
        push certificate password
``` 
but only the `.p12` file is actually supported.

It would be easy to load the certificate and key separately.
The `setupPushService` uses a helper to unbundle the `p12` file, but then the `push.NewClient` expects the unbundled x509.Certificate + rsa.PrivateKey.
Check to see if the config is not empty and load the cert and key from the given paths. 

For reference: 
https://golang.org/pkg/crypto/x509/#ParseCertificate
https://golang.org/pkg/crypto/x509/#ParsePKCS1PrivateKey
micromdm/certhelper has a bunch of loading/parsing code for keys. 

It would be perfectly acceptable to only support loading PEM formatted certs/keys. That's the format that you get the cert in from identity.apple.com and from certhelper.

PR #60, opened 2016-11-14
Fix for #55 providing ability to provide PEM cert & key for the APNS …

jessepeterson:
…cert (instead of only P12) (#55)


jessepeterson:
Please critique harshly. :)  This seems a bit hacky the way we're it's being done here by reading/decoding the certificate twice, essentially - once for the actual load for the APNS lib and the other for extracting the topic. Seems like once the cert is successfully decoded we ought to be using that. However I haven't extensively reviewed the architecture to know how best to solve that or if it would too tightly couple components so for now just duplicate some logic.

enroll/certificates.go
```diff
@@ -1,22 +1,39 @@
...
 )
 
-const PushTopicASN1 string = "0.9.2342.19200300.100.1.1"
+const (
+	certificatePEMBlockType string = "CERTIFICATE"
+	PushTopicASN1                  = "0.9.2342.19200300.100.1.1"
+)
 
-func GetPushTopicFromPKCS12(certPath string, certPass string) (string, error) {
+func GetPushTopicFromCert(certPath, certPass, keyPath string) (string, error) {
```
groob: Avoid using `Get...` in the name of the function. `TopicFromCert`, same as in buford looks perfectly clear. 

It should also be private `topicFromCert`, no reason to offer this in the public API, especially since we might update bufford and use that instead.

enroll/certificates.go
```diff
@@ -1,22 +1,39 @@
...
 	certData, err := ioutil.ReadFile(certPath)
 	if err != nil {
 		return "", err
 	}
 
-	_, cert, err := pkcs12.Decode(certData, certPass)
-	if err != nil {
-		return "", err
+	var cert *x509.Certificate
```
groob: I agree with you that it seems like we're doing double work here with decoding the same certs. If this service requires the cert and key, we can move this decoding to the setup code, and change the inputs to NewService to accept a *x509.Certificate and a *rsa.PrivateKey instead of accepting string paths.

app/services.go
```diff
@@ -155,15 +157,31 @@ func (s *serviceManager) setupPushService() {
...
+			s.APNS.CertificatePath,
+			s.APNS.PrivateKeyPass,
+		)
+		if err != nil {
+			s.err = err
+			return
+		}
+		tlscert = pushcertificate.TLS(cert, key)
+	} else {
+		tlscert, err = tls.LoadX509KeyPair(
```
groob: If you decide to move the certificate unmarshalling code from the enroll package here, you can just use `pushcertificate.TLS(...)` regardless of wether it's a PEM or a p12.

enroll/certificates.go
```diff
@@ -1,22 +1,39 @@
...
+	"encoding/pem"
 	"errors"
 	"golang.org/x/crypto/pkcs12"
 	"io/ioutil"
 )
 
-const PushTopicASN1 string = "0.9.2342.19200300.100.1.1"
+const (
+	certificatePEMBlockType string = "CERTIFICATE"
+	PushTopicASN1                  = "0.9.2342.19200300.100.1.1"
```
groob: The PushTopic could be changed to an oid var, and then you can use the correct `.Equals` function from `asn1`. 

The change would be something like `var oidASN1Topic = asn1.ObjectIdentifier{0,9,2342,...}` and then you can compare with https://golang.org/pkg/encoding/asn1/#ObjectIdentifier.Equal
I know the const was there from the previous PR, but I think it would be good to clean up :)

jessepeterson: Roger that.  The PKCS12/PushTopic code already existed and I minimally changed it. Noted nonetheless, though!

enroll/certificates.go
jessepeterson: Roger that.  The PKCS12/PushTopic code already existed and I minimally changed it. Noted nonetheless, though! I didn't catch that this was in the same package, so that can certainly change.

enroll/certificates.go
jessepeterson: Let's talk structure/architecture at some point. I'd like to see the capability for multiple MDM "servers" (that is push topics+URLs) supported under one single daemon/server.

groob: Can you open a short issue about this so we remember to discuss? (Maybe just link to your comment here). I've thought a bit about this very briefly as well.

enroll/certificates.go
```diff
@@ -11,7 +11,17 @@ import (
 
 const certificatePEMBlockType string = "CERTIFICATE"
 
-var oidASN1Topic = asn1.ObjectIdentifier{0, 9, 2342, 19200300, 100, 1, 1}
+var oidASN1UserID = asn1.ObjectIdentifier{0, 9, 2342, 19200300, 100, 1, 1}
```
groob: Nice attention to detail!

app/services.go
```diff
@@ -68,15 +77,82 @@ type serviceManager struct {
...
+			s.err = err
+			return
+		}
+		s.PushServiceCert.PrivateKey, s.PushServiceCert.Certificate, err =
+			pkcs12.Decode(pkcs12Data, s.APNS.PrivateKeyPass)
+		if err != nil {
+			s.err = err
+			return
+		}
+	} else {
```
groob: up to you, but I'd clean it up a bit by doing the following:

```
s.PushServiceCert.PrivateKey, s.PushServiceCert.Certificate, s.err = pkcs12.Decode(pkcs12Data, s.APNS.PrivateKeyPass)
return
```
the drop the `else`(since we return right above), and then do the same on line 123.

PR #63, opened 2016-11-20
add event publisher

groob:
Previously we were writing events directly to a datastore. This branch is the beginning of work to transition publishing events to a message queue instead. 

move previous checkin service to `checkin/service/direct`
add new event package for MDM events. use an NSQ producer to publish events to a topic.
add new implementation of checkin service in `checkin/service/event` The event version writes `Authenticate, TokenUpdate and Checkout` events to individual topics. 

Here's an example way to "subscribe" to an event topic with NSQ. This can be done outside the micromdm process. 

```
package main

import (
	"fmt"
	"log"
	"time"

	"github.com/nsqio/go-nsq"
)

func main() {

	cfg := nsq.NewConfig()

	c, err := nsq.NewConsumer("mdm.Authenticate", "printer", cfg)
	if err != nil {
		log.Fatal(err)
	}
	go func() {
		c.AddHandler(nsq.HandlerFunc(func(m *nsq.Message) error {
			fmt.Println(string(m.Body))
			return nil
		}))

		if err := c.ConnectToNSQD("localhost:4150"); err != nil {
			panic(err)
		}
	}()

	// Sleep a little to give everything time to start up and let
	// our producer and consumer run
	time.Sleep(5 * time.Minute)

}
```

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -142,6 +142,9 @@ func (c *Config) loadDEPConfig(enabled, sim bool, ck, cs, at, as, serverURL stri
 		AccessSecret:   as,
 		ServerURL:      serverURL,
 	}
+	if !sim && (ck == "" || cs == "" || at == "" || as == "") {
+		config.Enabled = false
+	}
 	if sim {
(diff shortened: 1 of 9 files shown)


groob:
Going to close this and push against new core branch.

PR #77, opened 2017-03-20
Fix dep config init

jessepeterson:


serve.go
```diff
@@ -488,7 +488,7 @@ func (c *config) depClient() (dep.Client, error) {
 			AccessSecret:   "AS_c31afd7a09691d83548489336e8ff1cb11b82b6bca13f793344496a556b1f4972eaff4dde6deb5ac9cf076fdfa97ec97699c34d515947b9cf9ed31c99dded6ba",
 		}
 	} else {
-		conf = &dep.Config{}
+		boltConf = new(dep.Config)
```
groob: Remove the else and declare it at the top of the bolt View func.

serve.go
```diff
@@ -507,6 +507,9 @@ func (c *config) depClient() (dep.Client, error) {
 		if err != nil {
 			return err
 		}
+		if !depsim {
+			conf = boltConf
+		}
```
groob: Instead of doing this check, assign every value to `boltConf` below, and then set `conf = boltConf`. if theres nothing in the bucket we would've bailed already.

serve.go
jessepeterson: The else is to avoid two potential initializations. Also bolt docs say to do as little as humanly possible inside the txn funcs so I init outside the txn. I'll defer to you, though.

serve.go
jessepeterson: Roger, this makes sense.

serve.go
groob: ok I think that makes sense.

PR #111, opened 2017-03-22
Fix panic where popFirst() can return nil

jessepeterson:
We were immediately using the nil pointer afterwards.


groob:
Well that's a fail on my part! This is why tests/code review should be rigorous :)

jessepeterson:
There is a test case (sorta). I sent an invalid InstallApplication command: literally just the UDID and the request_type (no manifest or anything).

Issue #85, opened 2017-03-20
provide consistent logging throughout the app


groob:
Right now we `fmt.Pritnf` or `log.Printf` where convenient, but that's not a sustainable strategy longterm. 

How logging should be implemented: 
Use the go-kit log package, it's great
https://github.com/go-kit/kit/tree/master/log

the log should provide a separate [context](https://github.com/go-kit/kit/tree/master/log#contextual-loggers) for each component and two logging levels configurable by the user: `debug`(a firehose of data useful for debugging/development) and `info` which will be the default logging mode. 

Labelling this as `beginner-friendly`. We'll have to think holistically about logging through the app, but pretty much anyone is welcome to take any `.go` file and improve logging by switching the fmt/stdlib logging to a go kit logger. :)

PR #112, opened 2017-03-22
Log all HTTP requests with gorilla handlers.CombinedLoggingHandler

jessepeterson:
This should also support #85.

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -32,6 +32,7 @@ import (
 	"github.com/go-kit/kit/endpoint"
 	"github.com/go-kit/kit/log"
 	httptransport "github.com/go-kit/kit/transport/http"
+	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


jessepeterson:
_Supports_ 85, but doesn't solve it — lots more to consistent logging than this. :)

groob:
This is a good start for sure 👍 

I'd like to eventually have all logging be key/value with https://github.com/go-kit/kit/tree/master/log#contextual-loggers

But the gorilla handlers are great for familiar http logs.

serve.go
```diff
@@ -177,6 +178,7 @@ func serve(args []string) error {
 	} else {
 		handler = r
 	}
+	handler = handlers.CombinedLoggingHandler(os.Stdout, handler)
```
groob: The other logger here goes to stderr
https://github.com/micromdm/micromdm/blob/472d416791f9ef822a6601763b0a6ffbc7c4b022/serve.go#L80

Thoughts on one or the other?

jessepeterson:
Yep. Part of the thought process is to help folks wrap heads around the typical MDM exchanges of enrollment, checkin, etc. Familiar-looking logs helps that. :)

serve.go
jessepeterson: Yeah, I saw that. Generally stdout = normal, stderr = errors. But for a long-running daemon those conventions are sorta.. I dunno.

Eventually of course the logging framework should be flexible enough to log to file, network, API, blah blah blah.

Issue #110, opened 2017-03-22
define user interface(CLI) to communicate with server


groob:
Old slack convos don't necessarily make for a great RFC, but this convo from october is a good refresher of where we'd like things to go :)
![screenshot 2017-03-22 00 04 35](https://cloud.githubusercontent.com/assets/1526945/24181747/6a65a230-0e93-11e7-8814-ad1d50ba1ba3.png)
![screenshot 2017-03-22 00 05 20](https://cloud.githubusercontent.com/assets/1526945/24181752/718284a2-0e93-11e7-8c0e-6f59e1300ff5.png)

PS: I really like that profile syntax.

Issue #82, opened 2017-03-20
add list/describe/get command to mange devices


groob:
Need to implement the following commands:
- `micromdm list devices` which will print out a list of devices with a format like below and some flags to filter devices. 
```
UDID     SerialNumber    EnrollmentStatus    LastSeen
```

- `micromdm describe devices -serial=SERIAL` 
Which will print a detailed view of a device. 

And finally:

- `micromdm get devices -serial=SERIAL -o=json` and get a JSON output of the device.

groob:
`get` was added in #129. 
`list` doesn't need to be there because get 1 vs get is just some filtering by udid/serial.

PR #114, opened 2017-03-22
[WIP] add list devices api

groob:
this is the beginning of #110 and #82.

Using the example here: https://github.com/go-kit/kit/tree/master/examples/addsvc

### DIFF
--- a/cmd/mdmctl/list.go
+++ b/cmd/mdmctl/list.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"text/tabwriter"
+
+	"github.com/go-kit/kit/log"
+	"github.com/micromdm/micromdm/core/list"
+)
+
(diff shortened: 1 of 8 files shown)


groob:
Lots of hacks in this PR. I ended up leaving grpc out of it because I didn't want to bind the server on multiple ports just to satisfy some endpoints. 
JSON over HTTP also feels like a simpler barrier to entry for most folks. 
This is the end result over http right now:
```
groob$ mdmctl list devices
UDID                                  SerialNumber  EnrollmentStatus  LastSeen
FA01680E-98CA-5557-8F59-7716ECFEE964  C026R7GXG7L2  true              0001-01-01 00:00:00 +0000 UTC
                                      C02ZD7GXW7R2  false             0001-01-01 00:00:00 +0000 UTC
```

I'll be merging this shortly and iterate from there on the `mdmctl` binary independently from the serverside APIs.

Issue #103, opened 2017-03-21
Add example syntax to -h command


loyaltyarm:
When using `sudo micromdm serve -h`, there is no example syntax given. This would be helpful to see outside of the specific example in the quickstart, simply because some commands use a `=` sign and some do not. Feel free to close if too trivial, but an example would be where a particular user is using `.pem` and not `.p12`.

groob:
`=` is optional for every flag. 

The way it works is `-flag=value` or `-flag value`. 
But there are defaults for each. 
For example, with a boolean value, you have the option of:

- omitting it: will default to whatever the `-h` flag says. 
- using the short version, `-flag`, which would set it to true, or
- with both the flag and the value `-flag false` or `-flag=false` 

I do think an `micromdm serve -examples` flag would be a really neat way to provide extra documentation for a command.

loyaltyarm:
Closing with #119

PR #119, opened 2017-03-23
add an -examples flag with some initial examples

loyaltyarm:
Addresses https://github.com/micromdm/micromdm/issues/103

This adds an `-examples` flag to `serve.go`. This is called separately from the `-h` command, so let me know how you feel about that and I'll change it accordingly. This is my first `Go` attempt so "go" easy on me. 😂

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -86,10 +86,16 @@ func serve(args []string) error {
 		flHTTPDebug    = flagset.Bool("http-debug", false, "enable debug for http(dumps full request)")
 		flRepoPath     = flagset.String("filerepo", "", "path to http file repo")
 		flDepSim       = flagset.Bool("depsim", false, "use depsim config")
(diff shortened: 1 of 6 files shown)


serve.go
```diff
@@ -86,12 +86,18 @@ func serve(args []string) error {
 		flHTTPDebug    = flagset.Bool("http-debug", false, "enable debug for http(dumps full request)")
 		flRepoPath     = flagset.String("filerepo", "", "path to http file repo")
 		flDepSim       = flagset.Bool("depsim", false, "use depsim config")
+		flExamples     = flagset.Bool("examples", true, "prints some example usage")
```
groob: This should be set to false, otherwise it will always print the examples.

loyaltyarm:
Sorry about all the commits! Won't click that button again!

Issue #100, opened 2017-03-21
Add job that fires APNS notifications automatically if there are commands queued


groob:
Need a worker which will 
a) ask the device to check in if a new commands was queued or a delayed one in case of NotNow. 
b) ask the device to check in every x hours (24 by default?)

jessepeterson:
@groob: what did you think of this note in #122 to address your concerns for _not_ wanting to send push notifications in some scenarios (which I agree with):

> One thing I was thinking was an attribute or some signal on each queued command that tells whether to send a push or not. In this way its up to the command-sender to decide when pushes get sent out (including some way to trigger an arbitrary push say at the end of queuing a bunch of commands).

This would allow _not_ sending pushs for queued commands when they land. Basically add a `PushRequested` to each command (defaulted to true). If this is done there may need to be a generic method to send a push notification (e.g. at the end of a group of queued commands).

Also: you'd said a device doesn't need push requests sent until after enrollment: if that's true we'll need to be able to queue new commands/behavior the moment a device enrolls (before the end of their first command) such that when the current command is done there are already more queued commands. Perhaps another event on a message bus to signal this? Not sure.

groob:
@jessepeterson I think the `PushRequested` idea merits some experimentation, but will need refinement. We'll definitely have to have something like it implemented in the long run though. 

Here's a somewhat hairy (but I expect very common) scenario to think through. 

* Device Enrolls, and we have a blueprint for the device which includes:
    - profileX
    - profileY
    - ApplicationA
    - UserAccount

* Device went through checkin process, and we queued all the commands to get that stuff from the blueprint. Now this device goes to `/connect` in a loop until we've replied with every single one of the queued commands. 
The device starts responding to use immediately with statuses:
ProfileX - Acknowledged 
ProfileY - Acknowledged
ApplicationA - NotNow - (we keep this in the queue for later)
UserAccount - Acknowledged
We respond with 200 OK and the device is done talking to us. 

... moments, minutes, hours etc later:
The user logs in. We somehow have to check in with the device and still deliver that InstallApplication command. At this point we need to start using the Push API to do it, because the device is not connecting on it's own.

The key here is that we want InstallApplication to be sent right after a user session started, but not just before, where we'll get a `NotNow`. 

Here's a log that matches the above scenario fairly closely:

[...]

PR #122, opened 2017-03-24
Send APNS push to UDID after Commands being queued.

jessepeterson:
In support of #100. Create new protobuf & message type for post-queued commands (QueuedCommand). The push service listens for these and sends an APNS push to the devices. Feedback very welcome. Note #100 will still need some sort of polling to check for devices that haven't checked in (as well as regular checkins, and NotNow handling).


jessepeterson:
One final though is how groups of commands might be orchestrated. I.e. if you wanted to queue and send 10 commands you only need the final Push notification. Currently the code would send 10 pushes. I imagine something a bit like some database abstractions use commits/rollbacks. Another method might be to simply delay the sending and not to "stack" multiple push notifications — but this terribly hacky.

push/service.go
```diff
@@ -2,20 +2,47 @@ package push
...
+	}
+	go func() {
+		for {
+			select {
+			case event := <-commandQueuedEvents:
+				udid, _, err := queue.UnmarshalQueuedCommand(event.Message)
+				if err != nil {
+					fmt.Println(err)
+				}
+				_, err = pushSvc.Push(nil, udid)
```
groob: One of the things I'd like this code to do is to _not_ send push notifications when a device enrolls. If the device is just enrolling it will continuously check in to the server until we return no more payloads. 
We'll want to handle pushing notifications to the device only after enrollment. 
Maybe each device get's it's own queue/subscription and we have Start() and Stop() methods on that queue?

Of course what I'm proposing is hard, so feel free to punt on that for now.

queue/command_queued.go
```diff
@@ -0,0 +1,21 @@
+package queue
+
+import (
+	"github.com/gogo/protobuf/proto"
+	"github.com/micromdm/micromdm/queue/internal/commandqueuedproto"
+)
+
+func MarshalQueuedCommand(udid, uuid string) ([]byte, error) {
```
groob: I'd prefer to keep Marshal, Unmarshal functions consistent across different types and packages. That means they should all have a signature like 

```
type Foo struct{}

MarshalFoo(*Foo) ([]byte, error)
UnmarhalFoo(data []byte, into *Foo) error
```

As long as we keep the methods consistent everywhere it will be easier to develop.

push/service.go
jessepeterson: One thing I was thinking was an attribute or some signal on each queued command that tells whether to send a push or not. In this way its up to the command-sender to decide when pushes get sent out (including some way to trigger an arbitrary push say at the end of queuing a bunch of commands).

queue/command_queued.go
jessepeterson: Fair enough. I only did it this way to avoid the proliferation of more little single-use structs.

queue/command_queued.go
```diff
@@ -5,17 +5,28 @@ import (
...
 )
 
-func MarshalQueuedCommand(udid, uuid string) ([]byte, error) {
+type QueueCommandQueued struct {
+	DeviceUDID  string
+	CommandUUID string
+}
+
+func MarshalQueuedCommand(cq *QueueCommandQueued) ([]byte, error) {
+	if cq == nil {
```
groob: This will cause a panic up from the caller. I would prefer 
`panic(you passed a nil thing here)` since it's a programmer error. 
or `return nil, fmt.Errorf(marshalling nil QueueCommmand)`

jessepeterson: I debated on that one. :) I'll do the latter.

PR #127, opened 2017-03-27
Add authentication to mdmctl and server

groob:
uses basic auth to set/read token

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -7,6 +7,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/asn1"
+	"encoding/base64"
 	"encoding/json"
 	"encoding/pem"
 	"flag"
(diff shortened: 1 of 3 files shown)


serve.go
```diff
@@ -721,3 +727,22 @@ func debugHTTPmiddleware(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+func basicAuth(password string) string {
```
jessepeterson: I think I'm missing where this function is used? It looks like you've used the http `SetAuthBasic()` and `AuthBasic()` funcs.

groob: This method is actually not used right now. I should remove it.

groob:
@jessepeterson the API credentials and server url for the client are set via environment variables right now but that's temporary. 

IMO on the client we should have a `~/.micromdm/` folder in the user's home directory to preserve config.

PR #133, opened 2017-03-29
Make ~/.micromdm if it does not exist

jessepeterson:
Also use more secure file mode bits for default.json.

### DIFF
--- a/cmd/mdmctl/config.go
+++ b/cmd/mdmctl/config.go
@@ -109,7 +109,7 @@ func SaveClientConfig(cfg *ClientConfig) error {
 	if err != nil {
 		return err
 	}
-	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return err
 	}
--- a/cmd/mdmctl/config.go
+++ b/cmd/mdmctl/config.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"os/user"
+	"path"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/mdmctl/config.go
```diff
@@ -80,21 +81,35 @@ func setCmd(cfg *ClientConfig, args []string) error {
...
 		return "", err
 	}
-	path := usr.HomeDir + "/.micromdm/default.json"
-	return path, nil
+	return path.Join(usr.HomeDir, ".micromdm"), nil
+}
+
+func clientConfigPath() (string, error) {
+	dir, err := clientConfigDirPath()
+	return path.Join(dir, "default.json"), err
```
groob: Why the extra method?

cmd/mdmctl/config.go
```diff
@@ -80,21 +81,35 @@ func setCmd(cfg *ClientConfig, args []string) error {
...
+	if err != nil {
+		return err
+	}
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		err = os.MkdirAll(path, 0777)
+		if err != nil {
+			return err
+		}
+	}
+	path, _ = clientConfigPath()
```
groob: Don't ignore err here

cmd/mdmctl/config.go
jessepeterson: Separate func to get just the directory (to make it). Alternatively could just return an []string and have the last user path.Join() them. I don't like post-parsing path strings.

cmd/mdmctl/config.go
jessepeterson: It's not possible, the way the functions are constructed, to get an error case here.

cmd/mdmctl/config.go
groob: Imo, use a single filepath.Join(homedir, .micromdm, default.json) and return that. 
When calling Mkdirall, use filepath.Dir(path) to get the dir.

cmd/mdmctl/config.go
jessepeterson: On second thought — even though that's true (there can never be an error here — it's probably best to check anyway, in case those functions get modified.

cmd/mdmctl/config.go
jessepeterson: Yeah I don't like `Dir()` - it's unnecessary string parsing because we already know all the path elements going in.

groob: ./shrug The extra call to path.Join with a wrapped func looks more awkward to me, but I'm fine with it if that's your preference.

cmd/mdmctl/config.go
groob: Sounds good to me. Feel free to merge if you update the err check here.

cmd/mdmctl/config.go
jessepeterson: Shrug as well. The efficiency gains are unimportant here. :) We'll go for cleanliness.

cmd/mdmctl/config.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"os/user"
+	"path"
```
groob: `path/filepath` has the same methods but looks like it would handle edge cases better.

Issue #118, opened 2017-03-23
Verify the device identity certificate


groob:
We want to verify the device identity certificate which was signed by the scep CA. 
This cert is only necessary for the `/mdm/connect` endpoint. Unfortunately cert auth happens during a TLS handshake so we can't verify it in the http handler. 

Here are the options for the connect endpoint:
Bind to `:8443` (configurable by the user) and require the identity certificate there. 

or: 

use `tls.Config.GetConfigForClient func(*ClientHelloInfo) (*Config, error)` and require certificate for 
`connect.mdm.acme.co`

Either way the user has to do additional configuration.

jessepeterson:
We can also go the MDM enrollment profile `SignMessage` route which passes the client identity over a `Mdm-Signature` header. (doc'd [here](https://developer.apple.com/library/content/documentation/Miscellaneous/Reference/MobileDeviceManagementProtocolRef/6-MDM_Best_Practices/MDM_Best_Practices.html#//apple_ref/doc/uid/TP40017387-CH5-SW1)).

PR #134, opened 2017-03-29
Implement SignMessage handling to verify client SCEP certificate (#118)

jessepeterson:


serve.go
```diff
@@ -728,6 +731,64 @@ func debugHTTPmiddleware(next http.Handler) http.Handler {
 	})
 }
 
+func mdmAuthSignMessageMiddleware(db *boltdepot.Depot, next http.Handler) http.HandlerFunc {
```
groob: Can you add a TODO that this should be moved to a library later?

serve.go
```diff
@@ -728,6 +731,64 @@ func debugHTTPmiddleware(next http.Handler) http.Handler {
...
+		cert := p7.GetOnlySigner()
+		if cert == nil {
+			http.Error(w, "Invalid signer", http.StatusBadRequest)
+			return
+		}
+
+		if hasCN, err := HasCN(db, cert.Subject.CommonName, 0, cert, false); err != nil {
+			fmt.Println(err)
+			http.Error(w, "Unable to validate signature", http.StatusInternalServerError)
+		} else if hasCN != true {
```
groob: Drop the else if here. Instead write
`hasCN, err := HasCN(...`
And `return` after each if. 

Also, `!hasCN` instead of != true

serve.go
```diff
@@ -746,3 +807,28 @@ func apiAuthMiddleware(token string, next http.Handler) http.HandlerFunc {
 
 	}
 }
+
+// implement HasCN function that belongs in micromdm/scep/depot/bolt
+//   note: added bool return, different from micromdm/scep interface
```
groob: Mind opening an issue about this note in micromdm/scep?

serve.go
jessepeterson: Good catch on the return; missed that!

I like the else there — where I can I like to keep variables scoped as small/short as possible and the bigger else statement achieves that. But I don't care enough to keep it that way. :) Yes, definitely `!hasCN`. That's what I get for late night coding!

serve.go
jessepeterson: Definitely. Actually I'm not sure about merging this until its implemented in micromdm/scep? What do you think? That way I can toss the bolt.Depot struct reference and instead use a scep.Depot interface in the config struct.

jessepeterson:
One note: if you're using micromdm you'll have to re-enroll all devices after this PR is merged. Non-SignMessage clients will be rejected (as they won't be sending Mdm-Signature headers).

PR #135, opened 2017-03-31
Avoid 500 response for empty command queue

jessepeterson:
Noticed that empty command queues were throwing a 500 error to clients. Catch that and handle a little more gracefully.

### DIFF
--- a/queue/queue.go
+++ b/queue/queue.go
@@ -27,6 +27,9 @@ type Store struct {
 func (db *Store) Next(ctx context.Context, resp mdm.Response) (*Command, error) {
 	dc, err := db.DeviceCommand(resp.UDID)
 	if err != nil {
+		if isNotFound(err) {
+			return nil, nil
+		}
 		return nil, errors.Wrapf(err, "get device command from queue, udid: %s", resp.UDID)
 	}
 
--- a/queue/queue_test.go
+++ b/queue/queue_test.go
@@ -137,7 +137,7 @@ func TestNext_zeroCommands(t *testing.T) {
 			resp := mdm.Response{CommandUUID: s, Status: s}
 			cmd, err := store.Next(ctx, resp)
 			if err != nil {
-				t.Error("expected nil but got err")
+				t.Errorf("expected nil but got err %s", err)
 			}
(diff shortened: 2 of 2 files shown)


queue/queue.go
```diff
@@ -27,6 +27,9 @@ type Store struct {
 func (db *Store) Next(ctx context.Context, resp mdm.Response) (*Command, error) {
 	dc, err := db.DeviceCommand(resp.UDID)
 	if err != nil {
+		if isNotFound(err) {
```
groob: `if err != nil && isNotFound(err {`

groob: actually this is fine, since there's other errors possible.

groob:
Need to also fix tests.

queue/queue.go
jessepeterson: Confused here. Are you proposing an `else if` here? Or two separate if statements the other of which catches the other error cases (the `Wrapf()` return)?

groob: I pushed submit before thinking. Ignore this comment

jessepeterson:
So you already said to fix tests, but I want to clarify here: the test is expecting Next to return an error for no commands but the command go-kit Service says `// next can return no errors and no payload.` So to be clear: that is acceptable — a no error, no payload response? If so I'll fixup the tests. If not I'll make `isNotFound` exportable and do the err test at after Next() usage (both in the tests and in the Service.

queue/queue.go
jessepeterson: I see your response now. I hadn't refreshed my page this morning so didn't see your reply. :)

queue/queue_test.go
```diff
@@ -136,8 +136,8 @@ func TestNext_zeroCommands(t *testing.T) {
 		t.Run(s, func(t *testing.T) {
 			resp := mdm.Response{CommandUUID: s, Status: s}
 			cmd, err := store.Next(ctx, resp)
-			if err == nil {
-				t.Error("expected err but got nil")
+			if err != nil {
+				t.Error("expected nil but got err")
```
groob: change this to errorf so we actually know what error was got

PR #137, opened 2017-04-03
Implement OTA

jessepeterson:
Beginning of implementation of [OTA configuration](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/iPhoneOTAConfiguration/Introduction/Introduction.html).


enroll/transport_http.go
```diff
@@ -23,10 +24,20 @@ func MakeHTTPHandlers(ctx context.Context, endpoints Endpoints, opts ...httptran
...
+			endpoints.OTAEnrollEndpoint,
+			nilRequest,
+			encodeResponse,
+			opts...,
+		),
 	}
 	return h
 }
 
+func nilRequest(_ context.Context, _ *http.Request) (interface{}, error) {
```
jessepeterson: @groob: gotta be a better way than this, yeah?

groob: No, that looks good. I have 
```
func decodeNoParamsRequest(ctx context.Context, r *http.Request) (interface{}, error) {
	return nil, nil
}
``` as an example I've done somewhere too.

**Comment by jessepeterson** [comment]
@groob: when you get a moment can you give this PR your thoughts? there's still some TODOs that might be worth pursuing (like actually encrypting the profile to the device's SCEP cert) but I think it's "functionally" there. i can enroll Macs devices using the `/ota/enroll` endpoint.

Note we're not really doing anything with the capabilities here (like the profile Challenge, or checking existing inventory for devices, using the provided attributes (like Serial, UDID, etc.) or anything like that. But the infrastructure is all there to do that at some point. Or for others to play with it for integration. This was sort of curiosity project.

**Comment by groob** [comment]
Hi @jessepeterson 

I took a look through the work you've done on this PR this morning. I think it looks great, and I'm fine with leaving the authentication and other parts unfinished. One thing you _could_ do is create a few issues around things that could deserve extra attention in future PRs, so that others can know where to look. 

I will have a share of nits on code sometime later today, but nothing that would affect where your PR is going.

**Comment by jessepeterson** [comment]
Roger that. I'll make issues for a few of the big things. Interested in your nits. I had particular concern about the best/cleanest place to separate access to various things (like the scepDepot getting shoved into the endpoint factories — mainly concerned with unnecessary coupling there). Thanks!

File: crypto/apple.go [hunk 2]
```diff
@@ -0,0 +1,99 @@
+package crypto
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+)
+
+const AppleRootCAPEM = `-----BEGIN CERTIFICATE-----
```
groob: Possible to keep these `const`s private?

enroll/endpoint.go
```diff
@@ -45,3 +82,44 @@ func MakeGetEnrollEndpoint(s Service) endpoint.Endpoint {
...
+		// given the inconsist certificate chain returned by macOS in OTA mode,
+		// macOS in DEP mode, and iOS in either mode. See:
+		// https://openradar.appspot.com/radar?id=4957320861712384
+		if err := crypto.VerifyFromAppleDeviceCA(req.p7.GetOnlySigner()); err == nil {
+			// signing certificate is signed by the Apple Device CA. this means
+			// we don't yet have a SCEP identity and thus are in Phase 2 of the
+			// OTA enrollment
+			profile, err := s.OTAPhase2(ctx)
+			return mdmEnrollResponse{profile, err}, nil
+		} else if caChain, _, err := scepDepot.CA(nil); err == nil && len(caChain) > 0 && req.p7.GetOnlySigner().CheckSignatureFrom(caChain[0]) == nil {
```
groob: I'm not a fan of eating the error here if it's not nil. Is it possible for us to do something else? 

I'd also like to see you get rid of `if -> else if` since you `return` in each branch.

enroll/profile.go
```diff
@@ -77,3 +77,18 @@ type MDMPayloadContent struct {
 	ServerURL               string
 	Topic                   string
 }
+
+var DefaultProfileServiceDeviceAttributes = []string{
```
groob: This list is only being accessed once, so let's move it inside the `OTAEnroll` method. We can revisit this in the future as needed.

enroll/service.go
```diff
@@ -8,6 +8,9 @@ import (
 
 type Service interface {
```
groob: Each one of these methods should have a comment in the form of 
```
// MethodName ...
```
That explains what the method does. It would be especially useful with the OTA methods since they're a bit cryptic to begin with.

enroll/transport_http.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/fullsailor/pkcs7"
+	httptransport "github.com/go-kit/kit/transport/http"
+	"github.com/groob/plist"
+)
+
+type HTTPHandlers struct {
+	EnrollHandler          http.Handler
+	OTAEnrollHandler       http.Handler
+	OTAPhase2Phase3Handler http.Handler
```
groob: Could you add a comment here explaining why Phase2 and 3 are in the same handler?

enroll/transport_http.go
```diff
@@ -0,0 +1,109 @@
...
+			return nil, err
+		}
+		return request, nil
+	default:
+		return nil, errors.New("unknown enrollment method")
+	}
+}
+
+func encodeResponse(ctx context.Context, w http.ResponseWriter, response interface{}) error {
+	switch resp := response.(type) {
```
groob: Let's drop the type switch here. IMO it's not necessary. 
If you'd like to keep it, I'd actually prefer a panic over an error here since it's a programmer error. 

Finally, instead of doing 
```
switch resp := ....
_ = resp
``` 
to avoid the unused variable error, you can do `switch resp.(type) {` without declaring a var.

enroll/transport_http.go
```diff
@@ -0,0 +1,109 @@
...
+	case mdmEnrollRequest, mdmOTAEnrollResponse:
+		_ = resp
+	default:
+		errors.New("unknown response type")
+	}
+
+	w.Header().Set("Content-Type", "application/x-apple-aspen-config")
+
+	if err := plist.NewEncoder(w).Encode(response); err != nil {
```
groob: Let's default to a pretty encoder (use https://godoc.org/github.com/groob/plist#Encoder.Indent) method here. 
I know the response goes to a device, but formatting will help with debugging.

serve.go
```diff
@@ -203,11 +203,13 @@ func serve(args []string) error {
...
 	scepHandler := scep.ServiceHandler(ctx, sm.scepService, httpLogger)
-	enrollHandler := enroll.ServiceHandler(ctx, sm.enrollService, httpLogger)
+	enrollHandlers := enroll.MakeHTTPHandlers(ctx, enroll.MakeServerEndpoints(sm.enrollService, sm.scepDepot), httptransport.ServerErrorLogger(httpLogger))
 	r := mux.NewRouter()
 	r.Handle("/mdm/checkin", mdmAuthSignMessageMiddleware(sm.scepDepot, checkinHandlers.CheckinHandler)).Methods("PUT")
 	r.Handle("/mdm/connect", mdmAuthSignMessageMiddleware(sm.scepDepot, connectHandlers.ConnectHandler)).Methods("PUT")
-	r.Handle("/mdm/enroll", enrollHandler).Methods("GET", "POST")
+	r.Handle("/mdm/enroll", enrollHandlers.EnrollHandler).Methods("GET", "POST")
+	r.Handle("/ota/enroll", enrollHandlers.OTAEnrollHandler)
+	r.Handle("/ota/phase23", enrollHandlers.OTAPhase2Phase3Handler).Methods("POST")
```
groob: Can you think of a more descriptive name than `phase23`? Maybe we can name this `/ota/finalize` ? or `/ota/start` and `/ota/complete`?

enroll/service.go
```diff
@@ -145,3 +148,57 @@ func (svc service) Enroll(ctx context.Context) (Profile, error) {
...
+	profile.PayloadOrganization = "MicroMDM"
+	profile.PayloadDisplayName = "Enrollment Profile"
+	profile.PayloadDescription = "The server may alter your settings"
+	profile.PayloadScope = "System"
+
+	scepContent := SCEPPayloadContent{
+		URL:      svc.SCEPURL,
+		Keysize:  1024,
+		KeyType:  "RSA",
+		KeyUsage: 0,
```
groob: is 0 even valid? 
https://golang.org/pkg/crypto/x509/#KeyUsage

I'd like to use a named const like `x509. KeyUsageDigitalSignature` if possible. 
If `0` happens to be the correct value maybe create a private const in this file?

enroll/service.go
```diff
@@ -145,3 +148,57 @@ func (svc service) Enroll(ctx context.Context) (Profile, error) {
...
+	profile := NewProfile()
+	profile.PayloadIdentifier = "com.github.micromdm.micromdm.ota-scep"
+	profile.PayloadOrganization = "MicroMDM"
+	profile.PayloadDisplayName = "Enrollment Profile"
+	profile.PayloadDescription = "The server may alter your settings"
+	profile.PayloadScope = "System"
+
+	scepContent := SCEPPayloadContent{
+		URL:      svc.SCEPURL,
+		Keysize:  1024,
```
groob: Can we use 2048?

enroll/service.go
```diff
@@ -145,3 +148,57 @@ func (svc service) Enroll(ctx context.Context) (Profile, error) {
...
+	scepPayload.PayloadContent = scepContent
+	scepPayload.PayloadScope = "System"
+
+	profile.PayloadContent = append(profile.PayloadContent, *scepPayload)
+
+	return *profile, nil
+}
+
+func (svc service) OTAPhase3(ctx context.Context) (Profile, error) {
+	return Profile{}, nil
```
groob: Do you need to add a `TODO: not implemented`? here?

enroll/endpoint.go
```diff
@@ -5,28 +5,65 @@ import (
...
+	Challenge     string `plist:"CHALLENGE"`
+	Product       string `plist:"PRODUCT"`
+	Serial        string `plist:"SERIAL"`
+	UDID          string `plist:"UDID"`
+	Version       string `plist:"VERSION"` // build no.
+	IMSI          string `plist:"IMSI"`
+	IMEI          string `plist:"IMEI,omitempty"`
+	MEID          string `plist:"MEID,omitempty"`
+	ICCID         string `plist:"ICCID"`
+	MacAddressEn0 string `plist:"MAC_ADDRESS_EN0"`
```
groob: `MACAddressEN0` is more idiomatic IMO. Initialisms should always be all caps.

enroll/endpoint.go
```diff
@@ -5,28 +5,65 @@ import (
...
+	GetEnrollEndpoint       endpoint.Endpoint
+	OTAEnrollEndpoint       endpoint.Endpoint
+	OTAPhase2Phase3Endpoint endpoint.Endpoint
 }
 
 type depEnrollmentRequest struct {
 	mdm.DEPEnrollmentRequest
 }
 
+// TODO: may overlap at some point with mdm.DEPEnrollmentRequest
```
groob: Do we need to create an issue in `micromdm/mdm`? Maybe we can find a more suitable name than `mdm.DEPEnrollmentRequest` which covers both uses. Something like:

```
type EnrollmentRequest struct {
    Serial        *string `plist:"SERIAL"`
}
```
Setting all fields as pointers would also allow using `omitempty` in the tags and checking if nil. 
In either case, I'd merge as is here but open an issue with the mdm package.

crypto/apple.go
```diff
@@ -0,0 +1,99 @@
...
+xptCVeN2pjbdt9uzi175oVo/u6B2ArKAW17u6XEHIdDMOe7cb33peVI6TD15W4MI
+pyQPbp8orlXe+tA8JA==
+-----END CERTIFICATE-----
+`
+
+// TODO: want to have more intensive verification
+func VerifyFromAppleDeviceCA(c *x509.Certificate) error {
+	block, _ := pem.Decode([]byte(AppleiPhoneDeviceCAPEM))
+	if block == nil || block.Type != "CERTIFICATE" {
+		return errors.New("invalid device CA")
```
groob: Should this be a `panic` since this error would never actually happen unless we update the PEM string and typo something?

The panic/error should read `device ca: invalid PEM block type %s",block.Type`

crypto/apple.go
```diff
@@ -0,0 +1,99 @@
...
+Z3pMViukVHe9WUg8Hum+0I/0kHKvjhwVd/IMwGlXyU7DhUYWdja2X/zqj7W24Aq5
+7dEKm3fqqxK5XCFVGY5HI0cRsdENyTP7lxSiiTRYj2mlPedheCn+k6T5y0U4Xr40
+FXwWb2nWqCF1AgIudhgvVbxlvqcxUm8Zz7yDeJ0JFovXQhyO5fLUHRLCQFssAbf8
+B4i8rYYsBUhYTspVJcxVpIIltkYpdIRSIARA49HNvKK4hzjzMS/OhKQpVKw+OCEZ
+xptCVeN2pjbdt9uzi175oVo/u6B2ArKAW17u6XEHIdDMOe7cb33peVI6TD15W4MI
+pyQPbp8orlXe+tA8JA==
+-----END CERTIFICATE-----
+`
+
+// TODO: want to have more intensive verification
```
groob: docstring should read 
```
// VerifyFromAppleDeviceCA ...
``` 

followed by the TODO at the bottom.

crypto/apple.go
jessepeterson: Not objection. Feels like, at some point, we'll need to provide access to them elsewhere? But perhaps Public funcs to return a set of proper structs? E.g. `func AppleRootCACertificate() *x509.Certificate () {}` or the like (which simply convert and return that?

That said, at the moment, we don't have a need for them to exported, so perhaps not a worry for now. I can change.

jessepeterson: Also: I want to kill the unused certs (just comment them out).

crypto/apple.go
jessepeterson: I left the error as "appleiPhoneDeviceCAPEM: invalid PEM block" as the whole PEM block could be nil, too.

jessepeterson: Also, panic'd on the other error.

enroll/endpoint.go
jessepeterson: Indeed. This is the messiest, hardest to read part of the whole thing. I was basically trying to cram everything onto one line in each test case. I knew this was going to draw ire. :)

There's two sets of operations on that line. One is retrieving the CA which is probably errorable, and the other is checking the certificate (which is probably worth ignoring if it doesn't match).

We do return in each branch, but there's an error case if neither branch is hit.

I can try and rejigger it.

enroll/service.go
jessepeterson: I'd love to. I merely copied it from the "normal" enrollment profile a few lines above so figured there was a reason it was used. As well Apple has this to say on it [here](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/iPhoneOTAConfiguration/profile-service/profile-service.html#//apple_ref/doc/uid/TP40009505-CH2-SW35):

> Key size requests the device to generate a keypair of a certain size. Only 1024-bit and 2048-bit key sizes should be used. Keys larger than 2048 bits are not supported. In general, 1024-bit keys are recommended because of the overhead involved in generating 2048-bit keys.

enroll/service.go
jessepeterson: Oops. Totally missed that! Though, it does work — nobody in the entire SCEP/OTA/Profile chain complained (including the scep server)!

The example docs use "5" as a bitfield for for digital signature and key encipherment. I'll check with profile docs reference, mosen's docs, and the Go consts to see if everything lines up.

jessepeterson: Also: again merely copied it from the enrollment profile above. But: the bit fields appear to match, can use constants.

Note the SCEP profile docs from Apple say:

> Some certificate authorities, such as Windows CA, support only encryption or signing, but not both at the same time.

serve.go
jessepeterson: Well, it neither starts, completes, or finalizes anything on its own. "StartAndFinalize" maybe? Its odd — the same URL is used for multiple steps (phases) in the OTA process. [This diagram](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/iPhoneOTAConfiguration/OTASecurity/OTASecurity.html#//apple_ref/doc/uid/TP40009505-CH3-SW1) is helpful but just keep in mind that phase 2 and 3 are both at the same URL.

Perhaps "enrollconfig" (Phase 2's name is "Device Enrollment" and Phase 3's name is "Device Configuration")?

Issue #138, opened 2017-04-05
arg -filerepo accepts non-existent paths


natewalck:
If you specify an invalid path for -filerepo, there is no indication it does not exist or you lack access to the path:

```
sudo $GOPATH/bin/micromdm serve -server-url=https://server.domain.com -apns-cert=/path/to/pushcert.p12 -apns-password="PASSWORD" -filerepo /foo/bar
```

It should exit saying the path is invalid rather than running as if it had worked.

PR #139, opened 2017-04-05
Add validation that filerepo path exists

natewalck:
Quick check to confirm filerepo path exists. Bail if it dosn't.

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -221,6 +221,9 @@ func serve(args []string) error {
 	}
 
 	if *flRepoPath != "" {
+		if _, err := os.Stat(*flRepoPath); os.IsNotExist(err) {
+			stdlog.Fatal(err)
+		}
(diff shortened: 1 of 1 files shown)


natewalck:
Quick check to "fix" #138

serve.go
```diff
@@ -221,7 +221,11 @@ func serve(args []string) error {
 	}
 
 	if *flRepoPath != "" {
-		r.PathPrefix("/repo/").Handler(http.StripPrefix("/repo/", http.FileServer(http.Dir(*flRepoPath))))
+		if _, err := os.Stat(*flRepoPath); err == nil {
```
groob: Use log.Fatal early, and drop the `else`. In Go it's generally better to avoid `else` in favor of early returns.

groob: You might also want to explicitly check for https://golang.org/pkg/os/#IsNotExist

PR #141, opened 2017-04-10
Move dep-token sub-command to mdmctl

jessepeterson:
Move the `dep-token` sub-command from a main micromdm CLI sub-command to be interfaced by `mdmctl`. Note this also starts storing new DEP token keypairs in the bolt DB config database bucket for DEP tokens (similar to SCEP). It will be used by mdmctl in the normal get/apply fashion by using the e.g. `mdmctl get dep-token` sub-commands (or the `/v1/dep-tokens` API endpoints).


cmd/mdmctl/apply.go
```diff
@@ -56,6 +60,7 @@ Apply a resource.
 Valid resource types:
 
   * blueprints
+  * dep-tokens
```
groob: I'd like to see the command itself be `dep`.

Imagine a syntax like:
```
mdmctl dep import-token /path/to/token-file.p7m
mdmctl dep export-token /path/to/stoken.json
mdmctl dep profiles (lists profiles)
mdmctl dep devices (list devices in DEP)
```

Thoughts? 
I think the `get/apply` commands are great but there's some value in separating DEP resources from mdm server ones. 

PS: I'm not opposed to the current interface either but wanted to bring it up for discussion.

jessepeterson: Was get/apply only supposed to be blueprints? I thought it was a general interface for "getting" and "applying" (putting) resources into the MDM server (regardless of resource type).

Yeah I could see this, not opposed. Might adjust it slightly:

```
mdmctl dep token -list # list an abreviated list of our tokens
mdmctl dep token -import /path/to/token.p7m # import a token encrypted to our private key
mdmcel dep token -import-json /path/to/token.json # import a token, JSON
mdmctl dep token -export-json /path/to/exported.json # export a token, JSON
mdmctl dep token -export-json /path/to/t.json -ck CK_534323234... # export this particular token
mdmctl dep token -export-cert /path/to/cert.pem # export DEP token public key
```

But totally flexible here.

One thing I want to make sure of, though, is that the DEP tokens are plural. Any given MDM server ought to be able to support multiple DEP tokens. Perhaps things can appear singular to an end-user, but the backend ought to be able to support (or be designed in a way to support) multiple DEP tokens.

jessepeterson: All that said any objection to finishing up the PR, merging and we can refactor later?

groob: sure we can do that. 👍 

I think get/apply should be generic for mdm server resources, but for DEP it makes sense to talk directly to apple, so I'm looking for a way to provide a separation between
`mdmctl get devices` and `mdmctl get dep-devices`

jessepeterson:
This PR is now functional. @groob do you want to take a look when you get a chance? Per the thread there's a few things to be decided on, though:

* Exact end-user interface (i.e. mdmctl CLI options)
* Moving some utility functions into a `dep-token` package (or not) and/or into `depsync`, etc.
* Removing from get(list)/apply APIs and into different packages.
* Move some other utility functions into the `crypto` package (already marked)

That said perhaps we move with what we have for now and tidy that up later.

Issue #142, opened 2017-04-13
update dep files to toml


groob:
https://sdboyer.io/dep-status/2017-04-10/

Living on the edge.

Issue #86, opened 2017-03-20
configurable authentication for enrollment endpoint


groob:
The enrollment URL  is accessible through two separate methods:

DEP devices send a `POST`:
- because of restrictions with DEP we can only issue a 401 with`WWW Authenticate` here. 
Support a basic auth here for now. 

Users send a `GET` to `/mdm/enroll`. We have more options here but it's also way too possible configurations. Let's do two for now(and implement others as requested). 
- basic auth - same as DEP
- google with Oauth2. 

This works for me and will likely be common enough for a large number of users so it's a good start.

loyaltyarm:
Any keys on getting started here? Is this to gate access to enrollment by first authenticating with an Oauth2 provider?

groob:
Yes that's correct.

If you're looking to work on this, I suggest adding a `--basic-auth` flag to get the hang of it. 
You can see how it's done in [micromdm/squirrel](https://github.com/micromdm/squirrel/blob/e0bca95d58925b0d3f961f509b69fd6fa5ad746d/cli/serve.go#L170-L181)

If you'd like to add oauth (only works with `GET`) you can use this library https://godoc.org/golang.org/x/oauth2

It has support for many providers, although the documentation assumes you've implemented oauth before. 
Luckily there's a pretty good tutorial here(specific to google OAuth): 
https://cloud.google.com/go/getting-started/authenticate-users

To use your own provider, you have to set the correct TOKEN_URL and AUTH_URL in your endpoint https://godoc.org/golang.org/x/oauth2#example-Config

np5:
I think the [over-the-air](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/iPhoneOTAConfiguration/OTASecurity/OTASecurity.html#//apple_ref/doc/uid/TP40009505-CH3-SW1) profile delivery has to be considered here. This would allow a third party app to take care of the user authentication. This app would have to build a [Profile Service Payload](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/iPhoneOTAConfiguration/profile-service/profile-service.html#//apple_ref/doc/uid/TP40009505-CH2-SW4) that will redirect the devices to micromdm.

The Service Profile contains a challenge that could be a signed nonce with a timestamp (micromdm and the 3rd party app sharing the signing key). Micromdm would then receive the same kind of information about the device as during a DEP enrollment, verify the nonce and the timestamp, and then proceed or not by answering with an MDM profile.

[...]

PR #143, opened 2017-04-14
Verify device certificate on DEP enrollment

jessepeterson:


enroll/transport_http.go
```diff
@@ -61,8 +62,18 @@ func decodeMDMEnrollRequest(_ context.Context, r *http.Request) (interface{}, er
 		if err != nil {
 			return nil, err
 		}
-		// TODO: We should verify but not currently possible. Apple
-		// does no provide a cert for the CA.
+		err = p7.Verify()
```
groob: Should we return a `401` to the client if verification fails here?

groob:
I see that CI will fail now. Lets solve #142 in a separate PR then rebase these once that's merged.

enroll/transport_http.go
jessepeterson: Until we have a solution for #86 then I'm not sure we should respond with 401. That will prompt the end-user to login to the MDM when in fact we should be rejecting the connection. _That said_ we should probably respond with a `4xx` error (maybe `400`? )as it is indeed a client-side/request issue.

PR #145, opened 2017-04-19
Blueprint apply & get/list

jessepeterson:
Start work on apply/get/list for blueprints. Note this PR is based off some work in the (as-yet-not-merged #141), though it should be fairly independent code-wise of it (just some conflict/merge concerns).

### DIFF
--- a/cmd/mdmctl/apply.go
+++ b/cmd/mdmctl/apply.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
@@ -11,7 +12,9 @@ import (
 
 	"github.com/go-kit/kit/log"
 	httptransport "github.com/go-kit/kit/transport/http"
+	"github.com/micromdm/micromdm/blueprint"
 	"github.com/micromdm/micromdm/core/apply"
+	uuid "github.com/satori/go.uuid"
 )
 
 type applyCommand struct {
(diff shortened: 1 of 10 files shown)


cmd/mdmctl/apply.go
```diff
@@ -1,12 +1,19 @@
...
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
 	"github.com/go-kit/kit/log"
+	"github.com/micromdm/micromdm/blueprint"
+	uuid "github.com/satori/go.uuid"
+
+	"encoding/json"
```
groob: Stdlib imports should come first

cmd/mdmctl/apply.go
```diff
@@ -68,9 +78,90 @@ Examples:
 
 func (cmd *applyCommand) applyBlueprint(args []string) error {
 	flagset := flag.NewFlagSet("blueprints", flag.ExitOnError)
+	var (
+		flBlueprintPath    = flagset.String("f", "", "filename of blueprint JSON to apply")
+		flNewBlueprintPath = flagset.String("n", "", "filename of new blueprint JSON to create")
```
groob: I think only the -f flag should exist. 
The second one is confusing. Is it the template?

core/list/service.go
```diff
@@ -35,3 +51,142 @@ func (svc *ListService) ListDevices(ctx context.Context, opt ListDevicesOption)
...
+	err := db.View(func(tx *bolt.Tx) error {
+		c := tx.Bucket([]byte(depTokenBucket)).Cursor()
+
+		prefix := []byte("CK_")
+		for k, v := c.Seek(prefix); k != nil && bytes.HasPrefix(k, prefix); k, v = c.Next() {
+			var depToken DEPToken
+			err := json.Unmarshal(v, &depToken)
+			if err != nil {
+				// TODO: log problematic DEP token, or remove altogether?
+				continue
```
groob: Let's log it.

core/list/service.go
```diff
@@ -35,3 +51,142 @@ func (svc *ListService) ListDevices(ctx context.Context, opt ListDevicesOption)
...
+func RandomCertificateSerialNumber() (*big.Int, error) {
+	limit := new(big.Int).Lsh(big.NewInt(1), 128)
+	return rand.Int(rand.Reader, limit)
+}
+
+// TODO: move into crypto package
+func SimpleSelfSignedRSAKeypair(cn string, days int) (key *rsa.PrivateKey, cert *x509.Certificate, err error) {
+	key, err = rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return
```
groob: Avoid using naked returns. Explicit is more readable.

cmd/mdmctl/apply.go
jessepeterson: It's a helper utility to quickly generate a new blueprint (which you can then edit to make a real one). It generetes a UUID, puts in the server URL, etc. — bunch of details the server already knows so a user can get on with managing things rather than sorting out what fields and format the blueprint JSON needs to be in.

I'm open to moving it or change semantics or whatever. I just think its helpful. Perhaps when other capabilities are added it will be unnecessary.

core/list/service.go
jessepeterson: Roger. FYI This code was in #141; its seen been squashed and rebased, though.

core/list/service.go
jessepeterson: Typically I totally agree and err (hah) that way usually, but in this case 1) the function(s) are very small, 2) the flow is easy to understand, 3) returning somewhat usable values in the case of error conditions is preserved without too much boilerplate and variable declaration. 

I can swap it out and make it C-style with the three variable declarations at the top, though.

cmd/mdmctl/apply.go
groob: I agree with you that it can be helpful, but I think `-n` vs `-f` is confusing. Can we rename the `-n` flag and help text to match the fact that we're generating a template for the user to edit? Maybe `-generate-blueprint`?

core/list/service.go
groob: You can name the returns but still reference them in the `return foo, bar, baz` line.

groob: screenshot 2017-04-19 22 11 38

From https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

jessepeterson: I guess my bar for the difference between "medium sized" and "handful of lines" is lower than that page suggests. :) I can add specific returns if that feels better for the project. :)

jessepeterson:
@groob This is functional for both apply and get now. Have a look when you're ready.

One thing that's missing is blueprint removal. Perhaps just `blueprint.DB.Save()` with an nil pointer? But probably best to make a `blueprint.DB.Delete()` or `.Remove()`.

cmd/mdmctl/apply.go
```diff
@@ -74,10 +77,63 @@ Examples:
...
+		newBlueprint.Profiles = []blueprint.Mobileconfig{blueprint.Mobileconfig([]byte("this should be a configuration profile"))}
+
+		enc := json.NewEncoder(newBlueprintFile)
+		enc.SetIndent("", "  ")
+		err = enc.Encode(newBlueprint)
+		if err != nil {
+			return err
+		}
+
+		err = newBlueprintFile.Close()
```
groob: I think you should have `defer newBlueprintFile.Close()` right after line 116. It's more readable, and unlikely to fail.

jessepeterson: Huh.. usually I use the defer idiom. Perhaps I was thinking that because this function is sort of multifunction that it was cleaner somehow? Or maybe code that's been since deleted had other reasons? Don't remember now. I'll change it. :)

PR #147, opened 2017-04-23
Allow self-signed certificates with configuration option in mdmctl

jessepeterson:


cmd/mdmctl/config.go
```diff
@@ -10,12 +10,27 @@ import (
...
 }
 
+// skipVerifyHTTPClient returns an *http.Client with InsecureSkipVerify set
+// to true for its TLS config. This allows self-signed SSL certificates.
+func skipVerifyHTTPClient(skipVerify bool) *http.Client {
+	if skipVerify {
+		tlsConfig := &tls.Config{InsecureSkipVerify: true}
+		transport := &http.Transport{TLSClientConfig: tlsConfig}
+		return &http.Client{Transport: transport}
+	} else {
```
groob: Drop the `else` because you return early in both branches.

PR #148, opened 2017-05-01
Past-due reorganization of some crypto helpers into the crypto package

jessepeterson:


crypto/helpers.go
```diff
@@ -0,0 +1,82 @@
...
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"os"
+	"time"
+)
+
+func RandomCertificateSerialNumber() (*big.Int, error) {
```
groob: How about `GenerateCertificateSerialNumber` instead of `Random...`

crypto/helpers.go
```diff
@@ -0,0 +1,82 @@
...
+	limit := new(big.Int).Lsh(big.NewInt(1), 128)
+	return rand.Int(rand.Reader, limit)
+}
+
+func SimpleSelfSignedRSAKeypair(cn string, days int) (key *rsa.PrivateKey, cert *x509.Certificate, err error) {
```
groob: Let's get rid of naked returns here.

crypto/helpers.go
jessepeterson: I think "random" should at least be in the name to indicate the source of the actual number. But I don't feel that strongly about it. I'll change it to Generate for now.

PR #149, opened 2017-05-01
Initial implementation of Blueprint triggering based on new Blueprint key "When"

jessepeterson:
Actually use and trigger Blueprint-based commands on MDM events. Major missing components are more "events" on which to base the Blueprints (only support one currently—on `TokenUpdate`) and of course the predicate engine which will make laser-focused targeting of (groups of) devices possible. See the many TODO notes in `listener.go` file for various ruminations.

I marked this PR as _discussion, proposal, and RFC_ because it needs some architectural consideration. It was a quick & dirty modification to think better about how workflows will work and not necessarily worthy of actual merging (though, if it works, it works, too).


blueprint/internal/blueprintproto/blueprint.proto
```diff
@@ -5,6 +5,7 @@ package blueprintproto;
 message Blueprint {
 	string uuid = 1;
 	string name = 2;
-	repeated string manifest_urls = 3;
-	repeated bytes mobileconfigs = 4;
+	string when = 3;
```
groob: This is a no-no with protobuf. You can add fields but never remove them or change the order.

groob:
I like the idea and a mechanism like this is needed. Here are my immediate concerns:

1) The point of a blueprint is to be declarative. It's what a device should look at in the end. Ideally the "When" is an implementation detail the user doesn't  concern them with. I'll concede that many of us want this kind of granularity though. 
2) `When` is not really self descriptive name for this. Maybe `ApplyAfter`? Not sure if it's any better. 
3) The triggers should not reference things like `TokenUpdate`. That's internal to the MDM protocol and apple might change the name or `mdmclient` behavior, causing unnecessary confusion for the user. We need a `micromdm` specific abstraction here.

jessepeterson:
Good feedback! Replies:

> 1. The point of a blueprint is to be declarative. It's what a device should look at in the end. Ideally the "When" is an implementation detail the user doesn't concern them with. I'll concede that many of us want this kind of granularity though.

In an ideal world where MDM tracks real state (like Puppet and Chef) I'd agree. Until then (or maybe never) I view Blueprints a bit like manifests — _including_ full `conditional_item` flexibility.

But here's a though-experiment question: if I have profiles/apps/commands that I only want sent on, say, initial DEP enrollment how do you think the user ought to be able to configure that? Or a hypothetical of, say, a profile that gets installed/removed on a schedule (for, say, a Kiosk mode or something). 

I.e. if the "desired state" really is temporal, or event driven, or predicated how should that be represented for the end-user to configure?

> 2. When is not really self descriptive name for this. Maybe ApplyAfter? Not sure if it's any better.

Long ago you actually suggested this field go away and be folded into the predicate engine as just another predicate keyword to test against. In this way the whole When and ApplyAfter naming goes away.

The said, until that functionality shows up, yeah I'm fine with ApplyAfter, or really any other appropriate name. Not whetted to 'when' at all.

> 3. The triggers should not reference things like TokenUpdate. That's internal to the MDM protocol and apple might change the name or mdmclient behavior, causing unnecessary confusion for the user. We need a micromdm specific abstraction here.

I'm fine with an abstraction. Maybe just "AfterEnrollment" or some such (which is, essentially, after a successful first TokenUpdate). However we've seen, and especially with DEP InstallApplication sets, that's its _extremely_ picky about when and how things are installed. I think this will necessitate exposing protocol level details to the user so they pinpoint where things need to go. Also I think if someone's chosen to use micromdm I think its fair to say they'll be at least familiar with how the MDM spec/protocol works.  micromdm is not, say, Jamf Pro. :) I don't think there's a need to try overly hard to hide implementation details behind abstractions (and on end-user level). Again more out of necessity of the shittiness of the MDM implementation then anything.

blueprint/internal/blueprintproto/blueprint.proto
jessepeterson: Ah, good to know!

loyaltyarm:
Throwing in my 2 pennies here.

>But here's a though-experiment question: if I have profiles/apps/commands that I only want sent on, say, initial DEP enrollment how do you think the user ought to be able to configure that? Or a hypothetical of, say, a profile that gets installed/removed on a schedule (for, say, a Kiosk mode or something).

For my own use cases, I would say a Enroll/Check-in/Always state for installation preference is required. I might have things I always want the device to have (certificates, settings/configurations/policies), or things that are only needed at enrollment (bootstrapping pieces for instance).

PR #156, opened 2017-05-14
Profile service/store, API, and mdmctl management

jessepeterson:


core/apply/endpoint.go
```diff
@@ -68,3 +97,4 @@ type depTokensResponse struct {
 
 func (r depTokensResponse) error() error { return r.Err }
 func (r blueprintResponse) error() error { return r.Err }
+func (r profileResponse) error() error   { return r.Err }
```
groob: the methods should be under the type declaration. let's not continue the trend here.

profile/profile.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/gogo/protobuf/proto"
+	"github.com/micromdm/micromdm/profile/internal/profileproto"
+)
+
+type Mobileconfig []byte
+
+type Profile struct {
+	Identifier string
+	Mobileconfig
```
groob: Any reason for embedding this new type rather than having it as a field?

profile/profile.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/gogo/protobuf/proto"
+	"github.com/micromdm/micromdm/profile/internal/profileproto"
+)
+
+type Mobileconfig []byte
+
+type Profile struct {
+	Identifier string
```
groob: What's the proposal for identifiers. If we're going to use them to GET/Update from the CLI then the identifier should be human friendly (imo).

core/apply/endpoint.go
jessepeterson: do you just mean physically move the them in the file so they're grouped next to (line-wise) with the type delcaration?

profile/profile.go
jessepeterson: this was actually something that needs to be changed. i had mistakenly though that config profiles, when encrypted, were entirely encrypted. if this was the case we needed to capture the identifier in an out-of-band way to reference (e.g. in blueprints). however i totally spaced that only the `PayloadContent` is encrypted in a Profile, so the `PayloadIdentifier` in a profile is _always_ accessible even for a signed profile (just need to CMS decode it).

the revised protobuf, for now would just be the mobileconfig. _however_ i envision it being extended to have "type" ENUM that tells whether it is a full configuration profile, a profile _payload_ only (e.g. OTA), an "interpreted" profile that should have variables replaced, and other forms. also other metadata might eventually be included like modified date and all that jazz. but again for now it really is just the mobileconfig data (and the e.g. Bolt keys can be inferred from the actual profile data).

core/apply/endpoint.go
groob: yep.

profile/profile.go
jessepeterson: On second (third? fourth?) thought — I think it also makes sense to leave them in both the protobuf and the Profile type. it's the primary identification method of a unique profile — both by devices and micromdm. that is its used often enough to keep the identifier around separately even if it is duplicated in the profile data itself.

However, that brings up the problem of synchronization. I think perhaps the best bet, at least for now, would be to double-check them before persisting or loading for the first time - try to make sure the profile type matches the identifier. this would save parsing the mobileconfig all the time, for example.

profile/profile.go
jessepeterson: Nope. Laziness at having to type Mobileconfig twice. :)

Issue #154, opened 2017-05-08
Move remaining configuration server subcommands to mdmctl


jessepeterson:
Remaining configuration server subcommands (e.g. for DEP configuration) should be moved to an mdmctl & RPC-based functionality. I.e. this is the code in `device.go`, `get.go`, and `apply.go`.

PR #160, opened 2017-05-21
create DEP commands for the list/apply services

groob:
for #154 

This PR creates the server endpoints for the `get dep` interactions:
- account info
- device details
- profiles

The `mdmctl` implementation will come in a separate PR.


groob:
Thanks for the reminder. I'm going to implement the apply command for defining a DEP profile as well, and then I'll delete the subcommands from the server binary.

PR #163, opened 2017-05-22
Blueprint profile integration

jessepeterson:


blueprint/listener.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	for _, p := range bp.ProfileIdentifiers {
+		foundProfile, err := db.profDB.ProfileById(p)
+		if err != nil {
+			if profile.IsNotFound(err) {
+				fmt.Printf("Profile ID %s in Blueprint %s does not exist\n", p, bp.Name)
+				continue
+			}
+			fmt.Println(errors.Wrap(err, "fetching profile"))
```
groob: use `fmt.Printf` for printing instead of errors.Wrap. The `Wrap` method is only useful to be able to inspect an error from a higher level, but you're not returning here.

jessepeterson: My intent was to add some context to the error being printed (for debugging purposes). But no worries, easily enough changed.

PR #167, opened 2017-05-23
Use the Profile service/store for the enrollment profile.

jessepeterson:
This allows the custom modification of the enrollment profile by the end user. However as a trade off we now essentially store a "static" enrollment profile. This is intended to change in the future when "templated" profiles are implemented where dynamic substring replacement (for example) is allowed in a Mobileconfig. This will allow for more dynamic-yet-custom profiles in the case of e.g. per-device SCEP challenges without having to have completely generated (and thus less flexible profiles).

### DIFF
--- a/enroll/service.go
+++ b/enroll/service.go
@@ -1,20 +1,49 @@
 package enroll
 
 import (
+	"bytes"
 	"crypto/x509"
-	"golang.org/x/net/context"
 	"io/ioutil"
 	"strings"
+
+	"github.com/groob/plist"
+	"golang.org/x/net/context"
+
+	"github.com/micromdm/micromdm/profile"
 )
 
+const EnrollmentProfileId string = "com.github.micromdm.micromdm.mdm"
+
 type Service interface {
-	Enroll(ctx context.Context) (Profile, error)
+	Enroll(ctx context.Context) (profile.Mobileconfig, error)
 	OTAEnroll(ctx context.Context) (Payload, error)
 	OTAPhase2(ctx context.Context) (Profile, error)
 	OTAPhase3(ctx context.Context) (Profile, error)
 }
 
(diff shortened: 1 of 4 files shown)


jessepeterson:
One note that #168 provided is that we probably ought to have a mechanism to delete profiles before or shortly after we merge this. The problem being that if someone needs to regenerate the enrollment profile for any reason this code isn't able to do that unless the enrollment profile has been removed from the profile store — and there's no UI to do that just yet.

enroll/service.go
```diff
@@ -170,19 +236,27 @@ func (svc service) OTAEnroll(ctx context.Context) (Payload, error) {
...
+	profile.PayloadDisplayName = "OTA Phase 2"
 	profile.PayloadScope = "System"
 
 	scepContent := SCEPPayloadContent{
-		URL:      svc.SCEPURL,
-		Keysize:  1024,
-		KeyType:  "RSA",
+		URL: svc.SCEPURL,
+
+		// TODO: OTA spec says: "In general, 1024-bit keys are recommended
```
groob: IMO if it doesn't break enrollment we can go with 2048.

enroll/service.go
```diff
@@ -49,15 +78,37 @@ func NewService(pushTopic, caCertPath, scepURL, scepChallenge, url, tlsCertPath,
...
 		TLSCert:       tlsCert,
-	}, nil
+		ProfileDB:     profileDB,
+	}
+
+	// look for existing enrollment profile, and save a new one if not found
+	_, err = svc.ProfileDB.ProfileById(EnrollmentProfileId)
+	if err != nil && profile.IsNotFound(err) {
+		enrollmentProfile, err := svc.MakeEnrollmentProfile()
+		if err != nil {
```
groob: if err != nil return nil, err is preferable to returning the service

PR #171, opened 2017-05-28
Put push and command APIs behind API authentication middleware

### LINKED ISSUE #157: Command and Push APIs ought to be authenticated
The `/push` and `/v1/commands` probably want to be put behind the API authentication or protected in some other way. Certainly the latter command API.

jessepeterson:
Resolves #157.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # v1.0.1 TBD
 
+* /push and /v1/commands API endpoints require API authentication #157
 * Add `mdmctl` binary for interacting with the server over API. #127
 * Save DEP cursor for use after restart. #109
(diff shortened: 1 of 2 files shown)


groob:
I realized we have a CHANGELOG we should stop neglecting before the new release. Mind updating that file to add this change? 

I'll submit a PR soon to update some of the other missing changes we've made in master. 

https://github.com/micromdm/micromdm/blob/master/CHANGELOG.md

jessepeterson:
Happy to. How do you propose doing versioning? Just append to the latest "unreleased" section (currently `v1.0.1 TBD`) and then rename that to the release header when finished?

groob:
That's my current plan. Before cutting a release I'll update the changelog file to add a new TBD section and make the current one final.

Issue #162, opened 2017-05-21
Consistent `mdmctl` CLI UI


jessepeterson:
A few of the `apply` and `get` commands are very similar but have different flags and calling conventions. To make these consistent throughout.

groob:
with apply I noticed that you created 
```
mdmctl apply blueprints -generate-blueprint /path/to/new_blueprint.json
```
which creates a template and saves it to a path.

With DEP profiles, I did something similar, but 

```
mdmctl apply dep-profiles -template
```
My version prints the template to stdout.

We should settle on one way to do this. 

Personally, I'd prefer if we print to stdout instead of asking for a file, because it's more flexible. I can always redirect output to a file if I need to, but many times I want to use the `-generate-blueprint` file just to remind myself of what the fields are, and creating a file is too much commitment.

jessepeterson:
Yep, I'm on board with stdout. However there's a number of options that _do_ generate files. Should it be across the board? E.g. `get blueprints -json` `get profiles -f`. Should those also be stdout? I could see it either way. However with blueprints and profiles I think the "saving" portion is implemented in the "listing" portion, so we'd have to resolve how output is done overall in those commands to support that.

Another option is to support using `-` as an argument to the filename to output as a way to dump to stdout.

groob:
IMO, flags like `-json` should also print to stdout. Another argument for doing so is that I can use a tool like `jq` (or even sed/awk etc) to filter output further. 

Also, that's how kubernetes does it:
```
kubectl get pods -o json
``` 
prints a big json blob to stdout.

jessepeterson:
Yep! Overall I'm fine with stdout output unless there's a good reason not to (e.g. non-encoded or excessive binary output, very large data, etc.).

jessepeterson:
One consideration for profiles is that they are blobs and may be CMS encoded. Some folks might be surprised to have binary data being output.

groob:
IMO the output for something like that should be b64 encoded.

jessepeterson:
Sure, but then it defeats some of the purpose of going to stdout in the first place. :) I.e. every time I wanted to look at a profile I don't want to have to worry about piping through `openssl base64 -d`. I suppose it could auto-detect if the first few bytes are `<?xml` and act accordingly, but then its somewhat non-deterministic.

jessepeterson:
Perhaps the CLI UX follows a set of rules, something like: 

[...]

PR #173, opened 2017-05-28
update generate blueprint command to print to stdout

groob:
For #162 

The updated command is:

```
mdmctl apply blueprints -template
{
  "uuid": "4a65c558-ecab-40e5-91ce-6e4a7f3ad318",
  "name": "exampleName",
  "install_application_manifest_urls": [
    "https://mdmdev.corp.kolide.co/repo/exampleAppManifest.plist"
  ],
  "profile_ids": [
    "com.example.my.profile"
  ],
  "apply_at": null
}
```


groob:
@jessepeterson once question I have is about 
`"apply_at": null`
What should the values there be? Can we update the example with the most common one?

jessepeterson:
Absolutely. I was thinking about that the other day. Only supported value right now is ["Enroll"]. That can totally be pre-populated.

PR #180, opened 2017-05-28
Remove profile integration

AnkurGel:


groob:
The reason CI fails is because it tries to use the `vendor` version of the mdm library. 

Try to run `dep ensure -update github.com/micromdm/mdm` which will update the Gopkg.lock file for the project to use the updated version of the library.

Issue #107, opened 2017-03-22
help the user with anchor certs if the server certificate is self signed


groob:
from the MDM reference:

> Optional. Array of strings. Each string should contain a DER-encoded certificate converted to Base64 encoding. If provided, these certificates are used as trusted anchor certificates when evaluating the trust of the connection to the MDM server URL. Otherwise, the built-in root certificates are used.

most users work with PEM certificates, so it would be great to provide some sort of helper for anchor certificates. Maybe the template for the DEP profile can auto-set them?

jessepeterson:
Not only for anchor certs (for DEP) but also as a trusted cert in the enrollment profile. Tangentially related to #95 as the TLS cert could be issue by the MicroMDM CA (and thus already trusted by the enrollment profile as it would be included there).

jessepeterson:
There's a couple different avenues to take on this:

1. In `cmd/mdmctl/apply_dep_profile.go` _template_ generation code we could use the HTTPS cert that mdmctl is connecting to and insert that into the template file. We could do this conditionally depending on if the `skip-verify` configuration parameter is specified, too. E.g. `apply dep-profiles -template -use-server-cert-as-anchor` (or just no additional flag, use the `skip-verify` to do it automatically.
2. We could add a separate switch to the _template_ generation code that points to an arbitrary PEM certificate which gets included in the generated template. E.g. `apply dep-profiles -template -anchor /path/cert.pem`.
3. We could add a switch to the actual profile _apply_ step (that contacts the API endpoint) to tell micromdm to include the server's current HTTPS certificate in the DEP profile on behalf of mdmctl. I.e. `apply dep-profiles -f profile.json -insert-server-anchor` or something. This is nice because mdmctl just need not worry about which cert to use.

Which seems most appealing? I like something that's more on the explicit side so 2 or 3 seem best. It could also be some combination of the above. One consideration in all of the above is handling the server-side of things if micromdm is running without TLS at all (e.g. behind a reverse proxy) then it may not know about which certificates to trust per se, in which case 2 would work best.

Thoughts welcome!

jessepeterson:
Sounds like we're going with a combo of 1 and 2 per Slack conversations.

PR #181, opened 2017-05-28
Add anchor certificate to DEP profile template

jessepeterson:
In support of #107

### DIFF
--- a/cmd/mdmctl/apply_dep_profile.go
+++ b/cmd/mdmctl/apply_dep_profile.go
@@ -2,12 +2,14 @@ package main
 
 import (
 	"context"
+	"crypto/x509"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
 
 	"github.com/micromdm/dep"
+	"github.com/micromdm/micromdm/crypto"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/mdmctl/apply_dep_profile.go
```diff
@@ -16,15 +18,23 @@ func (cmd *applyCommand) applyDEPProfile(args []string) error {
 	var (
 		flProfilePath = flagset.String("f", "", "filename of DEP profile to apply")
 		flTemplate    = flagset.Bool("template", false, "print a JSON example of a DEP profile")
+		flAnchorFile  = flagset.String("anchor", "", "filename of PEM certificate to add to trusted anchors")
```
groob: Help text should indicate that anchor expects a valid chain.

jessepeterson: Hmm, what do you mean a valid chain? In the case of a self-signed cert a single cert is the "valid chain." Happy to modify the text though. What do you suggest?

groob: I can generate a self signed root CA, and then an intermediary, and then sign a cert with that intermediary. The whole chain needs to be added, no?

jessepeterson: Hmm. Well, I haven't tested. I'd think you can just trust the leaf and that's good. I would also think you could just trust the CA (but could see how that wouldn't work, too). I know it works specifying a self-signed, though, but that doesn't address the question.

groob: I doubt that a leaf would work in this scenario. Usually trust requires a full chain.

jessepeterson:
Using new mdmctl DEP commands (thanks @groob!) have confirmed that generating a DEP profile template using `mdmctl apply dep-profiles -template -use-server-cert` works for specifying a single self-signed cert. DEP enrollment successfully completes for macOS. I'm sure that'll be good enough for most folks. But I'd like to test using self-signed CA with intermediates. The code doesn't currently support having more than one certificate added per each method (either `-anchor` or `-user-server-cert`) so that could be a problem if that doesn't work and folks are using something like that.

Issue #93, opened 2017-03-20
provide an automated way of generating appmanifeststs 


groob:
We already have the `appmanifest` utility from the tools repo, but we can do better here. 
Here's an idea for a possible user experience:

`micromdm import-pkg /path/to/file.pkg`
which would:
a) validate that the package is signed (we could also look into automated signing but that's extra tools) 
b) validate it's a distribution package. - go-xar can help. 
c) generate an appmanifest and copy the package and manifest automatically to be served by the built in fileserver.

erikng:
Sick!
________________________________

groob:
I added this to the yvr milestone, would be a good addition to `mdmctl` to complete the DEP bootstrap story.

groob:
The API and `mdmctl` should also have a `get apps` endpoint where the user can list apps or get an appmanifest to edit.

jessepeterson:
I've said it in Slack before but I think there's a case to be made to store app manifests in the MDM server. In a not dissimilar way to how we store profiles. There's a couple reasons I think this makes sense:

* The appmanifest can contain bundle-ID information for the application to be installed. This can be used (in a very limited case) to do some form of idempotency using the `InstalledApplicationList` MDM command. Again while it only applies to application/bundle IDs it may still have value.
* Having the server have control of the appmanifest will allow us to use some form of protection of the manifest file. I.e. we can do download-count limiting, or based on InstallApplication commands we can generate unique per-command-issue URLs to serve manifests. This prevents guessing or information leak security problems in the manifest.
* Generally having that data local could be advantageous
  * We could have a "verify" utility that downloads all the bits and manifest details (as an mdmclient would) to help an end-user
  * Details that could be used for a UI
  * And various other uses for just having that data local to the mdm server
  * While yes, the manifest could be cached ad-hoc it may be advantageous to keep it local.

Per above if we do it like the profile store folks can download and upload replacement manifests to customize to their specific needs using mdmctl.

PR #192, opened 2017-05-30
create appmanifest package

groob:
ported from tools/appmanifest. First step in solving #93, which will let users upload pkgs with `mdmctl`

### DIFF
--- a/appmanifest/appmanifest.go
+++ b/appmanifest/appmanifest.go
@@ -51,15 +51,36 @@ type File interface {
 	Size() int64
 }
 
+type Option func(*config)
+
+// WithMD5Size overrides the DefaultMD5Size when creating an AppManifest.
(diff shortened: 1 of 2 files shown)


appmanifest/appmanifest.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// File is an io.Reader which knows its size.
+type File interface {
+	io.Reader
+	Size() int64
+}
+
+// Create an AppManifest and write it to an io.Writer.
+func Create(file File, url string, md5Size int64) (*Manifest, error) {
```
groob: size should be an optional arg, defaulting to the const at the top of the file

groob: Could also use optional args to add info like shine

groob:
Merging this to work on client and server pr that uses the new package.

PR #193, opened 2017-05-30
Consistency of profiles and blueprints CLIs

jessepeterson:
Adjust profiles and blueprints CLIs for get/apply commands to have consistent argument names and usage conventions.


jessepeterson:
For #162

PR #199, opened 2017-05-31
mdmctl: support generating appmanifest

groob:
generate an appmanifest. validates and optionally signs packages.

For #93 

No support for uploads yet, but I'll handle that in another PR. 

The package validates signature but not if the pkg is a distribution package. This is a TODO, as I wasn't sure what the best mechanism for checking that information would be. 

Usage: 
```
USAGE
  mdmctl apply app [flags]

FLAGS
  -manifest -        path to an app manifest. optional,
                     will be created if file does not exist.
  -md5size 10485760  md5 hash size in bytes (optional)
  -pkg               path to a distribution pkg.
  -pkg-url           use custom pkg url
  -sign              sign package before importing, requires specifying a product ID (optional)
```

Example:
```
mdmctl apply app -pkg ~/Downloads/munkitools-2.8.2.2855.pkg -sign "Developer ID Installer: Victor Vrantchan (FooBarBaz)"

productsign: using timestamp authority for signature
productsign: signing product with identity "Developer ID Installer: Victor Vrantchan (FooBarBaz)" from keychain /Users/victor/Library/Keychains/login.keychain-db
productsign: adding certificate "Developer ID Certification Authority"
productsign: adding certificate "Apple Root CA"
productsign: Wrote signed product archive to /var/folders/02/k62z5c796fs5gm651jc610v80000gn/T/munkitools-2.8.2.2855.pkg
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>items</key>
    <array>
      <dict>
        <key>assets</key>
        <array>
          <dict>
            <key>kind</key>
            <string>software-package</string>
            <key>md5-size</key>
            <integer>3471762</integer>
            <key>md5s</key>
            <array>
              <string>08c3e7843f8b81fd44de6386619d0c39</string>
            </array>
            <key>url</key>
            <string>https:/mdm.acme.co/repo/munkitools-2.8.2.2855.pkg</string>
          </dict>
        </array>
      </dict>
    </array>
  </dict>
</plist>
```

### DIFF
--- a/cmd/mdmctl/apply.go
+++ b/cmd/mdmctl/apply.go
@@ -56,6 +56,8 @@ func (cmd *applyCommand) Run(args []string) error {
 		run = cmd.applyDEPProfile
 	case "profiles":
 		run = cmd.applyProfile
+	case "app":
+		run = cmd.applyApp
 	default:
 		cmd.Usage()
 		os.Exit(1)
@@ -73,6 +75,7 @@ Valid resource types:
   * profiles
   * dep-tokens
   * dep-profiles
+  * app
 
 Examples:
   # Apply a Blueprint.
--- a/cmd/mdmctl/apply_app.go
+++ b/cmd/mdmctl/apply_app.go
@@ -2,7 +2,10 @@ package main
 
 import (
 	"bytes"
+	"compress/zlib"
+	"encoding/binary"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"net/url"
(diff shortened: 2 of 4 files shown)


jessepeterson:
FYI here's Commandment's use of xar C binary to verify that its signed and to generate bundle IDs from the Distribution file. https://github.com/jessepeterson/commandment/blob/master/commandment/utils/app_manifest.py#L13-L94

jessepeterson:
One other thought: we should probably always default to merely warning about signed state. E.g. if a user wants to (some day) generate manifests for .ipa files for iOS we may not be able to easily (or at all) verify that.

PR #216, opened 2017-06-16
Command response JSON URL callback

jessepeterson:
Tested with simple Python Flask app.

### DIFF
--- a/queue/queue.go
+++ b/queue/queue.go
@@ -2,7 +2,9 @@
 package queue
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/boltdb/bolt"
@@ -11,6 +13,7 @@ import (
 
 	"github.com/micromdm/mdm"
 	"github.com/micromdm/micromdm/command"
+	"github.com/micromdm/micromdm/commandcallback"
 	"github.com/micromdm/micromdm/pubsub"
 )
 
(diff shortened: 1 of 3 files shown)


queue/queue.go
```diff
@@ -22,6 +25,18 @@ const (
 
 type Store struct {
 	*bolt.DB
+	pubsub.Publisher
+}
+
+func (db *Store) publishResponseJSON(resp mdm.Response) error {
+	buf := new(bytes.Buffer)
+	enc := json.NewEncoder(buf)
```
groob: Why are you converting something that came as a plist to json? The conversion is lossy between the two 

IMO you should just take the []byte coming from the http request to /connect and pubslish it there.

serve.go
```diff
@@ -501,6 +506,15 @@ type config struct {
 	err error
 }
 
+func (c *config) setupCommandCallback() {
```
groob: I'm fine with this as an experiment but I think that if you introduce a publisher, it should be swappable everywhere the `pubsub` interface is used. 

The idea is to take the existing channel implementation and expose that to multiple consumers, not to create a bunch of random plugins. 

I would suggest that you wrap the inmem pubsub that publishes each event to both the inmem queue and to the callback.

jessepeterson: Not everything on the queue is readily marshal-able to JSON. Most of the pubsub messages are protobufs and I specifically wanted to keep that away from (this particular) callback. I.e. submit command w/ JSON API, receive URL callback of JSON.

queue/queue.go
jessepeterson: I wanted to have a bit of API parity bewtween the command submit API and the command receive API. I.e. both of them JSON. The mdm.Response also contains other data that might be useful to processing a command (or at least for debugging).

I considered making a new /api/v1/rawcommand/<uuid> API where you submit actual raw Plist MDM commands that go directly on the queue (and you would indeed get raw Plist commands coming from the client on the callback).

Didn't go that way honestly because this way is much, much easier. :)

serve.go
jessepeterson: Also: I hear you about the multiple consumers, I get it. We could cut the pubsub out of it altogether and just callback directly from the queue package, but I do like the separation of duties when using the pubsub queue.

queue/queue.go
groob: hmm, imo with the `/connect` responses the value is that you get the raw feed directly from the device. 

Right now we don't actually have all the possible responses from mdm, so if you exposed the raw plist coming directly to `/connect` the user would probably be able to take better advantage of it.

groob:
Closing this, but I'm not opposed to the idea, just the implementation. 

I think we can create an HTTP middleware that optionally forwards everything to the user. I am going to open an issue to track.

PR #217, opened 2017-06-16
DO NOT MERGE: experiment

groob:
just an example of a pubsub wrapper that pushes messages to two destinations. A MultiPublisher of sorts.

### DIFF
--- a/pubsub/callback/callback.go
+++ b/pubsub/callback/callback.go
@@ -1,0 +1,41 @@
+package callback
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/micromdm/micromdm/pubsub"
+)
+
+type CallbackPublisher struct {
+	url  string
+	next pubsub.Publisher
+}
+
+func (p *CallbackPublisher) Publish(ctx context.Context, topic string, msg []byte) error {
+
+	// send to callback url
+	var buf bytes.Buffer
+	var event = struct {
+		Topic   string `json:"topic"`
+		Message []byte `json:"message"`
+	}{
+		Topic:   topic,
+		Message: msg,
+	}
+	// TODO error handling
+	json.NewEncoder(&buf).Encode(&event)
+	http.Post(p.url, "application/json", &buf)
+
+	// call the curent publisher as normal
+	return p.next.Publish(ctx, topic, msg)
+}
+
+func NewCallBackPublisher(next pubsub.Publisher, url string) *CallbackPublisher {
+	return &CallbackPublisher{
+		url:  url,
+		next: next,
+	}
+}


pubsub/callback/callback.go
```diff
@@ -0,0 +1,41 @@
...
+	next pubsub.Publisher
+}
+
+func (p *CallbackPublisher) Publish(ctx context.Context, topic string, msg []byte) error {
+
+	// send to callback url
+	var buf bytes.Buffer
+	var event = struct {
+		Topic   string `json:"topic"`
+		Message []byte `json:"message"`
```
groob: this should marshal anything on the message bus to json, including the proto. The consumer might have to choose to decode the data on their own.

pubsub/callback/callback.go
```diff
@@ -0,0 +1,41 @@
...
+	}{
+		Topic:   topic,
+		Message: msg,
```
jessepeterson: For a byte array being JSON'd that'll turn it into base64 data. I'd want to avoid that for the command callback. The thought there is you should only need an HTTP server and a JSON parser for the callback (which may or may not decode base64 for you, so best to avoid it) just as you only need a JSON encoder and an HTTP client to _submit_ commands.

Also most (all?) of the other events on the bus are protobuf which I don't know needs to be exposed for this particular callback. They _could_ be, but that's outside the scope of what I was thinking for this callback. I.e. dirt simple JSON (only) for command responses.

jessepeterson: _Otherwise_ yeah, this is fine for e.g. all the other events. There is no command response event so we still end up needing to code a new event to publish on the bus as a special case.

pubsub/callback/callback.go
jessepeterson: Well, it'll turn it into a b64 byte array. I'd want to avoid that. I'd want json elements for all the command response dictionaries and keys and whatnot (just like the command-submit API) for this _particular_ API. Otherwise you have to mess with decoding and other nonsense when all you really want is the command responses in JSON.

pubsub/callback/callback.go
groob: I was planning on adding response to the pubsub bus in general so I'm not against that. 

This publisher could filter out the messages for the callback and only publish the ones you care about.

groob: The one thing I still don't agree with is that JSON makes a better option here over plist. 

The reason being is that `micromdm/mdm` still has a long way to go before we have accurate structs for ALL responses. 
if we publish the raw plist as it comes to `/connect` there would be a greater value to user right now, even if their callback handler has to decode a plist.

groob: Given your requirements, my counter proposal would be to implement an HTTP middleware around the connect handler which forwards the request body from `/mdm/connect` to this callback url. It can default to just the raw body, but we can also add a flag for a json re-encode. 

IMO that would be the simplest solution that achieves all the desired goals with minimal disruption.

groob:
Closing this. Good discussion fodder for #216

PR #218, opened 2017-06-26
Allow signed profiles to be uploaded by checking for PKCS7

jessepeterson:


profile/profile.go
```diff
@@ -16,9 +17,22 @@ type payloadIdentifier struct {
 }
 
 func (mc *Mobileconfig) GetPayloadIdentifier() (string, error) {
+	mcBytes := *mc
+	if len(mcBytes) > 5 && string(mcBytes[0:5]) != "<?xml" {
+		p7, err := pkcs7.Parse(mcBytes)
+		if err != nil {
+			return "", errors.Wrapf(err, "Mobileconfig is not XML nor PKCS7 parseable")
+		}
+		err = p7.Verify()
```
groob: what CA does the call to Verify use?

profile/profile.go
```diff
@@ -16,9 +17,22 @@ type payloadIdentifier struct {
...
+		if err != nil {
+			return "", err
+		}
+		mcBytes = Mobileconfig(p7.Content)
+	}
+
 	// TODO: support CMS signed profiles
```
groob: remove the TODO?

profile/profile.go
jessepeterson: It doesn't verify any certificates. PKCS7 verify merely verifies the signature of the data is valid.

jessepeterson: https://github.com/fullsailor/pkcs7/blob/master/pkcs7.go#L208-L221

jessepeterson: This is similar in the OpenSSL world, too — verifying the signature cryptographically verifies against the data (but not the included certificates, or where they're from, or anything like that).

profile/profile.go
jessepeterson: oh lord. yes, sorry :)

profile/profile.go
groob: Makes sense. Thanks for the refresher on how this works:)

PR #225, opened 2017-09-02
Power Nap leads to infinite loop

nboelter:
The following was tested on macOS High Sierra 10.13 Developer Beta 9, but i think it could also affect stable OS versions.

When i had my MacBookPro12,1 in Power Nap and sent it multiple commands, i noticed that micromdm would keep on trying those commands indefinitely: 

```
connected udid=macbook type=, status=Idle
responded CommandUUID 1
connected udid=macbook type=, status=NotNow
responded CommandUUID 2
connected udid=macbook type=, status=NotNow
responded CommandUUID 1
connected udid=macbook type=, status=NotNow
responded CommandUUID 2
connected udid=macbook type=, status=NotNow
responded CommandUUID 1
connected udid=macbook type=, status=NotNow
responded CommandUUID 2
```

One simple solution would be not to send any commands when a Status: NotNow is received. However in general devices will Ack some commands, even when they refuse others with NotNow.

So i added a new Queue, commands that were refused with Status: NotNow get added to the new queue. Those will only get sent when the regular Commands queue is empty and the Status we received is not NotNow, to prevent infinite loops.

It now looks like this when the device is in Power Nap

```
connected udid=macbook type=, status=Idle
responded CommandUUID 1
connected udid=macbook type=, status=NotNow
responded CommandUUID 2
connected udid=macbook type=, status=NotNow
responded CommandUUID 3
connected udid=macbook type=, status=NotNow
empty response
```
Then nothing happens until the device is woken up:
```
connected udid=macbook type=, status=Idle
responded CommandUUID 1
connected udid=macbook type=, status=Acknowledged
responded CommandUUID 2
connected udid=macbook type=, status=Acknowledged
responded CommandUUID 3
connected udid=macbook type=, status=Acknowledged
empty response
```

I added a fairly basic test as well.


queue/queue.go
```diff
@@ -36,13 +36,14 @@ func (db *Store) Next(ctx context.Context, resp mdm.Response) (*Command, error)
...
-		// move down, send next
+		// We will try this command later when the device is not
+		// responding with NotNow
 		x, a := cut(dc.Commands, resp.CommandUUID)
 		dc.Commands = a
 		if x == nil {
 			break
 		}
-		dc.Commands = append(dc.Commands, *x)
+		dc.NotNow = append(dc.NotNow, *x)
```
groob: What does the device see in the response? Just 200 OK with no response?

Have you seen the `Handling a NotNow Response` in the documentation?
https://developer.apple.com/library/content/documentation/Miscellaneous/Reference/MobileDeviceManagementProtocolRef/3-MDM_Protocol/MDM_Protocol.html#//apple_ref/doc/uid/TP40017387-CH3-SW406

queue/queue.go
```diff
@@ -78,13 +79,15 @@ func (db *Store) Next(ctx context.Context, resp mdm.Response) (*Command, error)
 	}
 
 	// pop the first command from the queue and add it to the end.
+	// If the regular queue is empty, send a command that got
```
groob: Any thoughts on sending the NotNow command before the rest of the queue? In other words, "pick up where we left off", rather than push NotNow commands to the back of the line.

nboelter: That is a really good question. As long as the device stays connected to APNS and the internet all the time, it should not be a problem and is probably the preferred order. 

a) You want the last command sent to have a Status: NotNow response so that the device polls later.
b) When the device is sleeping, new commands should be preferred because those in the NotNow queue would probably be refused again. If you have both queues filled, it would take twice as much time to empty the Commands queue.
c) Because the device keeps clearing the Commands queue, when it wakes up it will mostly get the NotNow queue anyway.
d) Some of the commands that will never get refused with NotNow are rather high priority, which gives the NotNow queue a slightly lower priority via association.

That logic mostly breaks down if the device first sleeps and then shuts down, so that both queues are filled when it reconnects.

queue/queue.go
nboelter: If the Command returned by Next is nil, the response will be empty yes.

And yes this solution depends on the device polling later basically.

nboelter:
I think there should be a way to enforce the order of commands, but probably not in queue.go? Because you might have to do some complex Error handling if say the third command in a sequence gets refused with Status: Error etc.

jessepeterson:
Firstly can you verify this behavior on a non-beta release OS? I want to double-check that the client isn't doing something goofy here.

We had this debate a while back and favored not complicating the queue code and instead utilizing the device's own `NotNow` retry code. I think we're doing something wrong if that's not the case. I think I might see the error with our existing code on that, actually: we merely check if the next command was the last command responded to by NotNow (which would be true for a single queued cmd, but that clearly doesn't work for multiple NotNow queue'd responses, I think.

However there are major command queue improvements (and I'm struggling to find the discussion thread we had on it — perhaps it was in Slack) that we've had planned in the past. A few of those:

1. Supporting ordered commands where some set of commands are guaranteed to be delivered to the client in a particular order
1. Better supporting out-of-order commands (including better NotNow handling).
1. Support ending a command chain on a non-NotNow and having scheduled APNS notification re-start a queue if necessary (though this can probably also be accomplished by just re-trying NotNow'd commands after Ack'd commands and stopping)
1. Prioritized queue commands that might 'jump the queue' in certain scenarios (e.g. ASAP mobile device wipes in, say, a security sensitive situation)

So in all I guess I'm ambivalent about this particular change, but don't feel that strongly about it. Especially considering the command queue has some major improvements planned for the future.

nboelter:
I finally found a way to get an iOS device to respond with NotNow: Disable TouchID, lock it and send an InstallProfile command.

I reproduced the problem with iOS 10.3.3 (latest release version), it connected a little over 600 times in the first minute. That can't be good for the battery :)

groob:
I'm merging this. Thanks @nboelter, good work.

PR #269, opened 2017-10-27
add accountconfiguration flags to blueprint

### LINKED ISSUE #261: Blueprint should allow specifying if user is admin/if user creation should be suppressed
With the current implementation adding a user to the blueprint enables the following conditions

screen shot 2017-10-24 at 8 37 15 pm

The blueprint should allow an administrator to specify these flags.

groob:
Closes #261


groob:
This works pretty well in my testing.

PR #275, opened 2017-10-28
Added ability to store multiple servers and switch between them

natewalck:
There are a few bugs with this still, but I wanted to get it diff'd for comment sooner rather than later.

Known issues:

- If servers.json doesn't exist, it doesn't create one
- It does not migrate existing default.json yet


natewalck:
I think I mangled the logic a little for when no config exists. I'll poke at that a little more.

cmd/mdmctl/config.go
```diff
@@ -92,6 +89,9 @@ func setCmd(cfg *ClientConfig, args []string) error {
 		return err
 	}
 
+	var cfg *ServerConfig
```
groob: `cfg := new(ServerConfig)` does the same.

cmd/mdmctl/config.go
```diff
@@ -147,13 +161,31 @@ func SaveClientConfig(cfg *ClientConfig) error {
 		return err
 	}
 	defer f.Close()
+	enc := json.NewEncoder(f)
+	enc.SetIndent("", "  ")
+	return enc.Encode(clientCfg)
+}
 
+func SaveServerConfig(cfg *ServerConfig, name string) error {
```
groob: These functions can be private(lowercase).

cmd/mdmctl/config.go
```diff
@@ -147,13 +161,31 @@ func SaveClientConfig(cfg *ClientConfig) error {
...
-	enc.SetIndent("", "  ")
-	return enc.Encode(cfg)
+	clientCfg.Servers[name] = *cfg
+	return SaveClientConfig(clientCfg)
+}
+
+func SwitchServerConfig(name string) error {
+	clientCfg, err := LoadClientConfig()
+	if err != nil {
+		return nil
```
groob: return err?

natewalck:
I removed the ClientConfig initialization from the `Run` function as I thought it made more sense in the save function (if it doesn't already exist). 

Let me know if any of this is completely misguided ;)

groob:
I think the change make sense and it's consistent with other usage of the runCmd patter we use for cli subcommands. 

BTW It's inspired by this project here https://github.com/oklog/oklog/blob/f857583a70c345341d679b3f27aa542c8db70a21/cmd/oklog/main.go#L42

I'm good with merging this(Nice Work!!) but I would like to see the functions like LoadConfig/SaveConfig etc be private. This is a `main` package and it's a bit odd to see exported types in there.

groob:
One thing this commit still needs is to print that `switch` is a subcommand when you type `mdmctl config -h`

```
mdmctl config -h             

mdmctl config print
mdmctl config set -h
```

groob:
```
mdmctl config print                                                                                                                                                                             2017/10/28 17:35:17 open /Users/victor/.micromdm/servers.json: no such file or directory
```

It should either read the old style config or print a message about updating the old config.

groob:
I have looked through the PR but was not able to figure out how I can update my current configuration in `default.json`  into the new format. 
Can you describe what the upgrade path for an existing `mdmctl` user is?

natewalck:
There is no mechanism yet to convert default to servers. I forgot to add it
:)

I’ll add it probably tomorrow or Monday along with the other feedback.

groob:
Thanks Nate.

natewalck:
OK, I *think* this is done, could use a sanity check on the migration flow.

cmd/mdmctl/config.go
```diff
@@ -76,6 +78,49 @@ mdmctl config set -h
...
+	}
+	if _, err := os.Stat(configPath); !os.IsNotExist(err) {
+		cfgData, err := ioutil.ReadFile(configPath)
+		var serverCfg *ServerConfig
+		err = json.Unmarshal(cfgData, &serverCfg)
+		if err != nil {
+			fmt.Errorf("failed to unmarshal %s : %s", configPath, err)
+		}
+		fmt.Println("Found old style config. Migrate to new config?")
+		if askForConfirmation() {
```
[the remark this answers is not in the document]
natewalck: Stomp over it and delete? The only thing we must ask for is a new name for the config (or I can shove it in as "default", but then that would annoy me and make me edit the json manually)

natewalck:
Better? I think it should ask what to name the config at the bare minimum.

cmd/mdmctl/config.go
```diff
@@ -74,14 +73,44 @@ func (cmd *configCommand) Usage() error {
...
+		return err
+	}
+	if _, err := os.Stat(configPath); !os.IsNotExist(err) {
+		cfgData, err := ioutil.ReadFile(configPath)
+		var serverCfg *ServerConfig
+		err = json.Unmarshal(cfgData, &serverCfg)
+		if err != nil {
+			fmt.Errorf("failed to unmarshal %s : %s", configPath, err)
+		}
+		fmt.Println("Found old style config. Migrating to new config...")
```
groob: I meant no confirmation at all.

natewalck:
Mildly annoying, but we must run the check *only if* we are not already running the `migrate` command.

I couldn't think of a clean way to do it in the case itself.

groob:
@natewalck I like this a lot

```
mdmctl config print
Found old style config. You must migrate it to continue
Run `mdmctl config migrate -name=myconfig`

```

👍 

```
mdmctl config print
mdmctl config migrate -name=dev

{
  "active": "",
  "servers": {
    "dev": {
      "api_token": "secret",
      "server_url": "https://mdm.acme.co/",
      "skip_verify": false
    }
  }
}
```

My last remaining nit is that the name from `migrate` should be set as the active server, otherwise it's not complete

```
mdmctl get devices

UDID  SerialNumber  EnrollmentStatus  LastSeen
Get /v1/devices: unsupported protocol scheme ""
```

cmd/mdmctl/config.go
```diff
@@ -74,14 +77,51 @@ func (cmd *configCommand) Usage() error {
...
+	return nil
+}
+func migrateServerConfig(configName string) error {
+	configPath, err := namedClientConfigPath("default.json")
+	if err != nil {
+		return err
+	}
+	cfgData, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		log.Fatal(err)
```
groob: return err instead of fatal

cmd/mdmctl/config.go
```diff
@@ -74,14 +77,51 @@ func (cmd *configCommand) Usage() error {
...
+	}
+	var serverCfg *ServerConfig
+	err = json.Unmarshal(cfgData, &serverCfg)
+	if err != nil {
+		fmt.Errorf("failed to unmarshal %s : %s", configPath, err)
+	}
+	err = saveServerConfig(serverCfg, configName)
```
groob: err is not checked.

groob: usually, if the err is the _only_ return, I like to use `if err := f(); err != nil { return err }` style instead of checking the error on a new line.

cmd/mdmctl/config.go
```diff
@@ -74,14 +77,51 @@ func (cmd *configCommand) Usage() error {
...
+		return err
+	}
+	cfgData, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		log.Fatal(err)
+	}
+	var serverCfg *ServerConfig
+	err = json.Unmarshal(cfgData, &serverCfg)
+	if err != nil {
+		fmt.Errorf("failed to unmarshal %s : %s", configPath, err)
```
groob: fmt.Errorf does not print the error but creates one. This string will never be printed. 

I suggest changing it to `return errors.Wrapf(err, "failed to unmarshal %s", configPath)` which will return an error with some context.

natewalck:
I think this covers all comments.

groob:
```
mdmctl get devices
unable to load default config file: open /Users/victor/.micromdm/servers.json: no such file or directory
```

groob:
It does work after I switch, but any mdmctl command should catch the old config and ask for a migrate step.

cmd/mdmctl/config.go
```diff
@@ -37,16 +39,18 @@ func (cmd *configCommand) Run(args []string) error {
 		os.Exit(1)
 	}
 
-	var config *ClientConfig
-	if cfg, err := LoadClientConfig(); err == nil {
-		config = cfg
-	} else {
-		config = new(ClientConfig)
+	if strings.ToLower(args[0]) != "migrate" {
```
groob: move this to the root of the `mdmctl` command tree, not just under `config`

PR #314, opened 2017-11-09
Fix quirky Makefile ENV vars

### LINKED ISSUE #313: Makefile has no $GOPATH

Master branch @ `d8b044f`

`make deps`

Dependencies get resolved and downloaded.

### What did you see instead?

```
$ make deps

folder is clonded in the wrong place, copying to a Go Workspace
See: https://golang.org/doc/code.html#Workspaces
fatal: could not create leading directories of '/src/github.com/micromdm/micromdm': Permission denied
Makefile:41: recipe for target 'check-deps' failed
make: *** [check-deps] Error 128
```

---

I've checked to see if $GOPATH was defined, but:

```Makefile
all: build

.PHONY: build

ifndef ($(GOPATH))
	GOPATH = $(HOME)/go
endif

echo:
	echo "~: $(HOME)"
	echo "GP: $(GOPATH)"
```

```
$ make echo
echo "~: /home/<my-user>"
~: /home/<my-user>
echo "GP: "
GP: 
```

It seems the `ifndef` step is not working.

Beanow:
Fixes #313 

Because the whoami has no other dependencies, it's safe to move this up a few lines and fixes the strange quirk that `ifndef` isn't triggered while it should be.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ all: build
 
 .PHONY: build
 
+USER = $(shell whoami)
 ifndef ($(GOPATH))
 	GOPATH = $(HOME)/go
 endif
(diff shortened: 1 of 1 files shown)


Beanow:
Never mind, this does not fix, but clear the variable.

PR #315, opened 2017-11-09
Command response webhook

### LINKED ISSUE #223: forward all device responses to  a callback url
Opening this to track the proposal from #216. 

The gist of the issue: until we have a good implementation that exposes pubsub, it might be useful to tee all device responses to a remote url. This can be accomplished through an HTTP middleware around `/connect`. 

In #216 we briefly discussed if the forwarded messages should be JSON or PLIST. 
JSON is potentially more convenient for someone implementing the webhook. On the other hand PLIST will be more accurate because that's what the device already sends.
I think we can parametrize the middleware such that both are accepted.

Beanow:
Closes #223 see #216 
A few notes:

As the marshalled object does not handle all possible content the raw data is passed along over pubsub as well. Making this webhook the first subscriber on this topic.

The `application/x-apple-aspen-mdm` is what I got from a real device. As such that's what I'm forwarding to the webhook.

The signature is no longer in the header, therefore it would be wise to include some type of secret in your webhook URL and/or set a trusted IP to authenticate the micromdm instance.


connect/connect.go
```diff
@@ -18,7 +18,7 @@ type ConnectService interface {
 
 	// Acknowledge acknowledges a response sent by a device and returns
 	// the next payload if one is available.
-	Acknowledge(ctx context.Context, req mdm.Response) (payload []byte, err error)
+	Acknowledge(ctx context.Context, req mdmConnectRequest) (payload []byte, err error)
```
groob: `mdmConnectRequest` is a private type and can't be used in the function signature. 
It's also a endpoint specific wrapper type, not intended to be exposed. 

I'd be more willing to consider a type like 

```type Response struct {
    Raw []byte
    MDMResponse mdm.Response
}
````

File: webhook/command_response.go [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/micromdm/micromdm/connect"
+	"github.com/micromdm/micromdm/pubsub"
+	"github.com/pkg/errors"
+)
+
+const contentType = "application/x-apple-aspen-mdm"
+
+type CommandResponse struct {
+	CallbackUrl string
```
groob: URL is an initialism and must be capitalized.

webhook/webhook.go
```diff
@@ -0,0 +1,9 @@
+package webhook
+
+import (
+	"github.com/micromdm/micromdm/pubsub"
+)
+
+type Webhook interface {
```
groob: This interface definition is not necessary.

webhook/command_response.go
```diff
@@ -0,0 +1,54 @@
...
+			case event := <-connectEvents:
+				var ev connect.Event
+				if err := connect.UnmarshalEvent(event.Message, &ev); err != nil {
+					fmt.Println(err)
+					continue
+				}
+
+				fmt.Printf("Content of event %s: %d\n", ev.Response.Status, len(ev.OriginalResponse))
+
+				c := http.Client{}
```
groob: By creating a new httpClient each time, you'll open a new file descriptor and run out of resources very fast. client should be a dependency of the `CommandResponse` struct, initialized in `package main` and passed down to all resources that require an HTTP Client.

webhook/command_response.go
```diff
@@ -0,0 +1,54 @@
...
+	"net/http"
+
+	"github.com/micromdm/micromdm/connect"
+	"github.com/micromdm/micromdm/pubsub"
+	"github.com/pkg/errors"
+)
+
+const contentType = "application/x-apple-aspen-mdm"
+
+type CommandResponse struct {
```
groob: It would be nice to keep this struct type a bit more generic so that other subscriptions can be added and forwarded (for example Checkin events like TokenUpdate and Checkout)

connect/transport_http.go
```diff
@@ -30,8 +32,10 @@ type errorer interface {
 }
 
 func decodeRequest(ctx context.Context, r *http.Request) (interface{}, error) {
-	var req mdmConnectRequest
-	err := plist.NewDecoder(r.Body).Decode(&req)
+	var res mdm.Response
+	body, err := ioutil.ReadAll(r.Body)
```
groob: err needs to be checked with `if err != nil`

webhook/command_response.go
Beanow: I'm making assumptions about the event type and how to unmarshal it, so I'm not sure how to cleanly generalize in this case.

Beanow:
Refactored to the best of my ability. Not often working with golang. Still not entirely sure how I could both avoid the interface and make things more generic. So feel free to edit that.

connect/transport_http.go
```diff
@@ -30,8 +32,19 @@ type errorer interface {
...
 func decodeRequest(ctx context.Context, r *http.Request) (interface{}, error) {
-	var req mdmConnectRequest
-	err := plist.NewDecoder(r.Body).Decode(&req)
+	var res mdm.Response
+
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return nil, err
+	}
+
```
groob: needs to have `defer r.Body.Close()` to avoid a resource leak

connect/transport_http.go
```diff
@@ -30,8 +32,19 @@ type errorer interface {
 }
 
 func decodeRequest(ctx context.Context, r *http.Request) (interface{}, error) {
-	var req mdmConnectRequest
-	err := plist.NewDecoder(r.Body).Decode(&req)
+	var res mdm.Response
+
+	body, err := ioutil.ReadAll(r.Body)
```
groob: reading the whole body into memory can be a security/DDOS problem. 
Change this line to the following
```
body, err := ioutil.ReadAll(io.LimitReader( r.Body,10000))
```

LimitReader: https://golang.org/pkg/io/#LimitReader

connect/transport_http.go
```diff
@@ -30,8 +32,19 @@ type errorer interface {
...
+		return nil, err
+	}
+
+	err = plist.Unmarshal(body, &res)
+	if err != nil {
+		return nil, err
+	}
+
+	req := MDMConnectRequest{MDMResponse: res, Raw: body}
 	return req, err
```
groob: the err was already returned, above, so this line can be nil.

serve.go
```diff
@@ -131,6 +133,9 @@ func serve(args []string) error {
 		APNSPrivateKeyPath:  *flAPNSKeyPath,
 		depsim:              *flDepSim,
 		tlsCertPath:         *flTLSCert,
+		CommandWebhookURL:   *flCommandWebhookURL,
+
+		webhooksHTTPClient: http.Client{},
```
groob: use `http.DefaultClient` instead.

serve.go
```diff
@@ -493,6 +502,9 @@ type config struct {
 	commandService command.Service
 	configService  configsvc.Service
 
+	responseWebhook    webhook.ListeningWebhook
+	webhooksHTTPClient http.Client
```
groob: the type should be a pointer `*http.Client`

webhook/command_response.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/micromdm/micromdm/connect"
+	"github.com/micromdm/micromdm/pubsub"
+	"github.com/pkg/errors"
+)
+
+const contentType = "application/x-apple-aspen-mdm"
+
+type commandWebhook struct {
+	Topic string
+	CallbackUrl string
```
groob: `s/CallbackUrl/CallbackURL`

webhook/command_response.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func NewCommandWebhook(topic, callbackURL string) (ListeningWebhook, error) {
+	if topic == "" {
+		return nil, errors.New("topic should not be empty")
```
groob: change the new errors to say `webhook: <error message>`
for example: `"webhook: topic should not be empty"`

otherwise it becomes hard to track where the error is coming from.

webhook/command_response.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	if callbackURL == "" {
+		return nil, errors.New("callbackURL should not be empty")
+	}
+	
+	return commandWebhook{Topic: topic, CallbackUrl: callbackURL}, nil
+}
+
+func (cw commandWebhook) StartListener(sub pubsub.Subscriber, httpClient http.Client) error {
```
groob: client should be a field on `cw` instead of a param.  `httpClient *http.Client`

webhook/webhook.go
```diff
@@ -0,0 +1,11 @@
+package webhook
+
+import (
+	"net/http"
+
+	"github.com/micromdm/micromdm/pubsub"
+)
+
+type ListeningWebhook interface {
```
groob: remove this interface and used an exported struct first. We can define the appropriate interface later when it becomes required.

webhook/command_response.go
```diff
@@ -0,0 +1,60 @@
...
+	"net/http"
+
+	"github.com/micromdm/micromdm/connect"
+	"github.com/micromdm/micromdm/pubsub"
+	"github.com/pkg/errors"
+)
+
+const contentType = "application/x-apple-aspen-mdm"
+
+type commandWebhook struct {
```
groob: make this `type Webhook struct` and keep it exported.

webhook/command_response.go
```diff
@@ -0,0 +1,60 @@
...
+	Topic string
+	CallbackUrl string
+}
+
+func NewCommandWebhook(topic, callbackURL string) (ListeningWebhook, error) {
```
groob: `NewCommandWebhook` should return an exported `Webhook` instead of the interface.

connect/transport_http.go
Beanow: I agree about the DDOS concern, but I think this is not the correct location to handle this. It should be done in a far earlier middleware.

For example, the `/connect` endpoint has a preceding middleware `mdmAuthSignMessageMiddleware` to check signatures and gets the body in the same way: `bodyBuf, err := ioutil.ReadAll(r.Body)`

Running 30MB through the hashing algorithm will probably have more of a DOS impact than the plist parsing would.

Beanow: Also, 10KB may be fairly constrained for some bodies, like the application list. Typically in other projects I go with 1MB for API endpoints that are not file uploads of some kind.

serve.go
Beanow: Actually looking for docs on this global instance I ran into this https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

So a new client with a timeout may be better in this case.

groob:
Merged as is, but going to do some additional follow-up in master. Thanks a lot for all your work @Beanow

Issue #311, opened 2017-11-07
Official Docker image


Beanow:
~Is there one? And if not, can we provide it?~

Edit: can we re-implement Docker builds with official releases being tagged?

Beanow:
I see there is something here: https://hub.docker.com/r/micromdm/micromdm/tags/
However it's outdated, has no tagged releases and it's Dockerfile does not show up in this repo.

groob:
This needs to be added. Thanks for filing an issue.

groob:
Closed by #316 

I will push a release up soon.

PR #316, opened 2017-11-09
Created a Dockerfile

Beanow:
See #311 

Example usage:

```bash
$ docker run --rm -ti -p 9000:443 \
	-v micromdm-config:/var/db/micromdm \
	-v $(pwd)/.certs:/.certs \
	micromdm/micromdm:latest serve \
		-apns-cert /.certs/apns.p12 \
		-apns-password NotPublicStuff \
		-api-key VerySecretThings \
		-server-url https://your-mdm-domain:9000 \
		-tls-cert="/.certs/tls-cert.pem" \
		-tls-key="/.certs/tls-private.pem"
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM alpine:3.6
+
+ENV MICROMDM_VERSION=v1.2.0
+RUN apk --no-cache add curl unzip ca-certificates && \
+	update-ca-certificates && \
+    curl -L https://github.com/micromdm/micromdm/releases/download/${MICROMDM_VERSION}/micromdm_${MICROMDM_VERSION}.zip -o /micromdm.zip && \
+    unzip -p /micromdm.zip build/linux/micromdm > /micromdm && \
+    rm /micromdm.zip && \
+    chmod a+x /micromdm && \
+    apk del curl unzip && \
+    echo '666e5457d396d84b1fc995886c7598741a487b50132f95305b6be44b20c7d834  micromdm' | sha256sum -c || exit 0
+
+EXPOSE 443
+VOLUME ["/var/db/micromdm"]
+ENTRYPOINT ["/micromdm"]
+CMD ["serve", "-examples"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,10 @@
 FROM alpine:3.6
 
-ENV MICROMDM_VERSION=v1.2.0
-RUN apk --no-cache add curl unzip ca-certificates && \
-	update-ca-certificates && \
-    curl -L https://github.com/micromdm/micromdm/releases/download/${MICROMDM_VERSION}/micromdm_${MICROMDM_VERSION}.zip -o /micromdm.zip && \
-    unzip -p /micromdm.zip build/linux/micromdm > /micromdm && \
-    rm /micromdm.zip && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    unzip -p /micromdm.zip build/linux/micromdm > /micromdm && \
+    rm /micromdm.zip && \
+    chmod a+x /micromdm && \
+    apk del curl unzip && \
+    echo '666e5457d396d84b1fc995886c7598741a487b50132f95305b6be44b20c7d834  micromdm' | sha256sum -c || exit 0
+
+EXPOSE 443
+VOLUME ["/var/db/micromdm"]
+ENTRYPOINT ["/micromdm"]
+CMD ["serve", "-examples"]
```
groob: CMD should be `micromdm serve` 

All the correct configuration should be passed through env vars

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    curl -L https://github.com/micromdm/micromdm/releases/download/${MICROMDM_VERSION}/micromdm_${MICROMDM_VERSION}.zip -o /micromdm.zip && \
+    unzip -p /micromdm.zip build/linux/micromdm > /micromdm && \
+    rm /micromdm.zip && \
+    chmod a+x /micromdm && \
+    apk del curl unzip && \
+    echo '666e5457d396d84b1fc995886c7598741a487b50132f95305b6be44b20c7d834  micromdm' | sha256sum -c || exit 0
+
+EXPOSE 443
+VOLUME ["/var/db/micromdm"]
+ENTRYPOINT ["/micromdm"]
```
groob: Leave out entrypoint

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM alpine:3.6
+
+ENV MICROMDM_VERSION=v1.2.0
+RUN apk --no-cache add curl unzip ca-certificates && \
```
groob: Build the dockerfile by cross compiling a linux dependency and using COPY to add the binary to the server, instead of pulling a release. 

Here's a good example of a Dockerfile I've written for another go project with similar startup/configuration:

https://github.com/kolide/fleet/blob/78b831a6d25f9242953b17b7e40b69aee3f698bf/Dockerfile

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+	update-ca-certificates && \
+    curl -L https://github.com/micromdm/micromdm/releases/download/${MICROMDM_VERSION}/micromdm_${MICROMDM_VERSION}.zip -o /micromdm.zip && \
+    unzip -p /micromdm.zip build/linux/micromdm > /micromdm && \
+    rm /micromdm.zip && \
+    chmod a+x /micromdm && \
+    apk del curl unzip && \
+    echo '666e5457d396d84b1fc995886c7598741a487b50132f95305b6be44b20c7d834  micromdm' | sha256sum -c || exit 0
+
+EXPOSE 443
```
groob: expose 80 for redirect to 443 as well

groob:
> @groob about using ENV vars for the configuration, while I think that's a nice addition, overhauling the config system to accept every argument as an ENV var too is a bit beyond the scope of my contribution.


Agreed about that. I just looked back because I thought I was already doing that.

PR #318, opened 2017-11-12
Support supplied depsim URL

jessepeterson:


serve.go
```diff
@@ -755,9 +755,8 @@ func (c *config) depClient() (dep.Client, error) {
 	}
 
 	depServerURL := "https://mdmenrollment.apple.com"
-	if depsim {
-		// TODO: support supplied depsim URL
-		depServerURL = "http://dep.micromdm.io:9000"
+	if depsim != "" {
```
groob: Why the same check twice? If it can be checked right above I'd move the assignment there.

jessepeterson: No reason. Just replaced the existing code. I can rearrange.

PR #326, opened 2017-11-19
DEP auto-assignment

jessepeterson:
See the associated 1.3 milestone for more details on this task

### DIFF
--- a/depsync/depsync.go
+++ b/depsync/depsync.go
@@ -19,10 +19,11 @@ import (
 const (
 	SyncTopic    = "mdm.DepSync"
 	ConfigBucket = "mdm.DEPConfig"
+	syncDuration = 30 * time.Minute
 )
 
 type Syncer interface {
-	privateDEPSyncer() bool
+	SyncNow()
 }
 
 type watcher struct {
@@ -33,6 +34,7 @@ type watcher struct {
 	publisher pubsub.Publisher
 	conf      *config
 	startSync chan bool
+	syncNow   chan bool
 }
 
 type cursor struct {
@@ -78,6 +80,7 @@ func New(pub pubsub.PublishSubscriber, db *bolt.DB, logger log.Logger, opts ...O
 		publisher: pub,
 		conf:      conf,
 		startSync: make(chan bool),
+		syncNow:   make(chan bool),
 	}
 
 	// apply our supplied options
@@ -148,9 +151,8 @@ func (w *watcher) updateClient(pubsub pubsub.Subscriber) error {
 	return nil
 }
 
-// TODO this is private temporarily until the interface can be defined
-func (w *watcher) privateDEPSyncer() bool {
-	return true
+func (w *watcher) SyncNow() {
+	w.syncNow <- true
 }
 
 // TODO this needs to be a proper error in the micromdm/dep package.
@@ -163,7 +165,7 @@ func isCursorExpired(err error) bool {
 }
 
 func (w *watcher) Run() error {
-	ticker := time.NewTicker(30 * time.Minute).C
+	ticker := time.NewTicker(syncDuration).C
 FETCH:
 	for {
 		resp, err := w.client.FetchDevices(dep.Limit(100), dep.Cursor(w.conf.Cursor.Value))
@@ -217,7 +219,11 @@ SYNC:
 			}
 		}
 		if !resp.MoreToFollow {
-			<-ticker
+			select {
+			case <-ticker:
+			case <-w.syncNow:
+				level.Info(w.logger).Log("msg", "explicit DEP sync requested")
+			}
 		}
 	}
 }
--- a/depsync/depsync.go
+++ b/depsync/depsync.go
@@ -26,6 +26,11 @@ type Syncer interface {
 	SyncNow()
 }
 
+type autoassigner struct {
+	filter       string
+	profile_uuid string
+}
+
 type watcher struct {
 	mtx    sync.RWMutex
 	logger log.Logger
@@ -35,6 +40,8 @@ type watcher struct {
 	conf      *config
 	startSync chan bool
 	syncNow   chan bool
+
+	assigners []autoassigner
 }
 
 type cursor struct {
@@ -81,6 +88,9 @@ func New(pub pubsub.PublishSubscriber, db *bolt.DB, logger log.Logger, opts ...O
 		conf:      conf,
 		startSync: make(chan bool),
 		syncNow:   make(chan bool),
+
+		// TODO: hard coded right now for initial testing
+		assigners: []autoassigner{autoassigner{filter: "*", profile_uuid: "95124484B114B710"}},
 	}
 
 	// apply our supplied options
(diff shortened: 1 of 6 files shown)


jessepeterson:
Note that as of right now there's still a bunch of work to do. Notably:

* mdmctl/core API to be able to dynamically set auto-assigners
* persistence layer to save and load the auto-assigners
* ability to re-assign the existing base of machines in the device DB to a DEP profile
  * probably needs some discussion on de-coupling of the device and depsync services/modules

depsync/depsync.go
```diff
@@ -71,33 +87,43 @@ func New(pub pubsub.PublishSubscriber, db *bolt.DB, opts ...Option) (Syncer, err
...
+		assigners: []autoassigner{autoassigner{filter: "*", profile_uuid: "95124484B114B710"}},
 	}
 
+	// apply our supplied options
 	for _, opt := range opts {
 		opt(sync)
 	}
 
+	// if no logger option has been set use the null logger
+	if sync.logger == nil {
```
groob: Add it before `opt(sync)` is called and then you don't need this check.

depsync/depsync.go
```diff
@@ -149,8 +174,57 @@ func isCursorExpired(err error) bool {
...
+			resultCounts[result] = resultCounts[result] + 1
+		}
+		// TODO: alternate strategy is to log all failed devices
+		// TODO: handle/requeue failed devices?
+		level.Info(w.logger).Log(
+			"msg", "dep-assigned",
+			"profile_uuid", profile_uuid,
+			"success", resultCounts["SUCCESS"],
+			"not_accessible", resultCounts["NOT_ACCESSIBLE"],
+			"failed", resultCounts["FAILED"])
```
groob: comma here + newline for the `)`

depsync/endpoint.go
```diff
@@ -0,0 +1,20 @@
+package depsync
+
+import (
+	"context"
+
+	"github.com/go-kit/kit/endpoint"
+)
+
+type Endpoints struct {
+	SyncNowEndpoint endpoint.Endpoint
```
groob: Maybe it could be called `Refresh` ? I don't mind this too much but it stutters a bit.

depsync/service.go
```diff
@@ -0,0 +1,22 @@
...
+type syncNowService struct {
+	syncer Syncer
+}
+
+func (s *syncNowService) SyncNow(_ context.Context) {
+	s.syncer.SyncNow()
+	return
+}
+
+func NewRPC(syncer Syncer) *syncNowService {
```
groob: Return an exported service, otherwise the types are hard to deal with.

serve.go
```diff
@@ -767,24 +771,27 @@ func (c *config) depClient() (dep.Client, error) {
...
 	if client != nil {
-		opts = append(opts, depsync.WithClient(client))
+		opts = append(opts, depsync.WithClient(client), depsync.WithLogger(logger))
 	}
-	_, c.err = depsync.New(c.pubclient, c.db, opts...)
+	var syncer depsync.Syncer
+	syncer, c.err = depsync.New(c.pubclient, c.db, logger, opts...)
 	if err != nil {
-		return
+		return nil
```
groob: log this err before returning nil.

depsync/depsync.go
```diff
@@ -159,18 +233,26 @@ FETCH:
...
-		if err != nil {
-			return err
-		}
-		if err := w.publisher.Publish(context.TODO(), SyncTopic, data); err != nil {
-			return err
+		if len(resp.Devices) > 0 {
+			e := NewEvent(resp.Devices)
+			data, err := MarshalEvent(e)
+			if err != nil {
+				return err
```
groob: errors.Wrap(err, .. somecontext) would be great here.

depsync/depsync.go
```diff
@@ -159,18 +233,26 @@ FETCH:
...
+			}
+			if err := w.publisher.Publish(context.TODO(), SyncTopic, data); err != nil {
+				return err
+			}
+			go func() {
```
groob: Any reason ths is async?

depsync/depsync.go
```diff
@@ -149,8 +174,57 @@ func isCursorExpired(err error) bool {
 	return strings.Contains(err.Error(), "EXPIRED_CURSOR")
 }
 
+func (w *watcher) processAutoAssign(devices []dep.Device) error {
```
groob: A lot of branching/relative complexity around these methods and state. Do you think this could be simplified further or maybe some of the logic could be extracted into functions?

depsync/depsync.go
```diff
@@ -202,9 +282,19 @@ SYNC:
...
+					level.Info(w.logger).Log("err", err, "auto-assign")
+				}
+			}()
 		}
 		if !resp.MoreToFollow {
-			<-ticker
+			select {
+			case <-ticker:
+			case <-w.syncNow:
+				level.Info(w.logger).Log("msg", "explicit DEP sync requested")
```
groob: why is there no action here? is this a stub?

depsync/depsync.go
```diff
@@ -159,18 +233,26 @@ FETCH:
 		} else if err != nil {
 			return err
 		}
-		fmt.Printf("more=%v, cursor=%s, fetched=%v\n", resp.MoreToFollow, resp.Cursor, resp.FetchedUntil)
+		level.Info(w.logger).Log("msg", "DEP fetch", "more", resp.MoreToFollow, "cursor", resp.Cursor, "fetched", resp.FetchedUntil, "devices", len(resp.Devices))
```
groob: use newlines to make this more readable. 

```
level.Info(w.logger).Log(
"msg", "DEP fetch",
"more", resp.MoreToFollow,
)
```
 keep at key, value per line

depsync/depsync.go
jessepeterson: But then I waste the resources of assigning/initializing a resource unnecessarily, no?

depsync/depsync.go
jessepeterson: No action? The point of the select (and the unconditional wait before it) is merely to wait for the ticker to expire or the SyncNow signal to fire to return early to do another sync again.

depsync/depsync.go
jessepeterson: Hmm, I suppose I could break separate the actual assigner logic from the DEP message processing.

depsync/depsync.go
jessepeterson: Don't want to hold up our DEP fetching to wait for all the auto-assign logic (and subsequent assign API calls).

depsync/depsync.go
jessepeterson: 👍

depsync/endpoint.go
jessepeterson: ¯\_(ツ)_/¯   sure. :)

depsync/service.go
jessepeterson: I modeled this on other existing services. Happy to change it.

serve.go
jessepeterson: It is logged, actually. The `c.err` is assigned to the config object and handled elsewhere. Most of the `setup*()` are odd like this. That _said_ there is a bug where I should have been checking `c.err` and not `err`!

depsync/depsync.go
jessepeterson: 👍

depsync/depsync.go
jessepeterson: 👍

depsync/depsync.go
```diff
@@ -19,19 +19,30 @@ import (
...
 )
 
 type Syncer interface {
-	privateDEPSyncer() bool
+	SyncNow()
+}
+
+type autoassigner struct {
+	filter       string
+	profile_uuid string
```
groob: use camel case `profileUUID`

jessepeterson: d'0h, good catch! autoassigner ought to be camelcased, too.

PR #360, opened 2017-12-14
fix: Get devices (and most other successful commands) should output to Stdout, not Stderr

### LINKED ISSUE #356: Get devices (and most other successful commands) should output to Stdout, not Stderr

v1.2.0 from releases page.

mdmctl get devices

Output on Stdout (to be able to further pipe command)

### What did you see instead?

Output on Stderr (making 'standard' piping a PITA)

Code here: https://github.com/micromdm/micromdm/blob/master/cmd/mdmctl/get.go#L126

thepriefy:
Get devices (and most other successful commands) should output to Stdout, not Stderr

Closes #356


groob:
Hi! Thanks for the fix!

Logger output should still go to street, but the tabwriter changes are exactly what we need. 

Mind changing the logger config back to Stdout?

thepriefy:
@groob I've edited logger config back to Stdout. Could you please check? 
Thank you. ^ ^

Issue #322, opened 2017-11-16
add the ability to filter serials 


groob:
```
mdmctl get device -h
```

```
  # Get a device by serial (TODO implement filtering)
  mdmctl get devices -serial=C02ABCDEF
```

this todo needs to be done.

jessepeterson:
Resolved by #363.

PR #363, opened 2018-01-01
search a serial out of all serials

bdemetris:
this is a first pass at allowing the ability to filter a single serial out of the full list of serials (devices), and tries to solve https://github.com/micromdm/micromdm/issues/322

### DIFF
--- a/cmd/mdmctl/get.go
+++ b/cmd/mdmctl/get.go
@@ -119,6 +119,9 @@ func (out *devicesTableOutput) BasicFooter() {
 
 func (cmd *getCommand) getDevices(args []string) error {
 	flagset := flag.NewFlagSet("devices", flag.ExitOnError)
+	var (
+		flFilterSerial = flagset.String("serial", "", "serial number by which to filter")
+	)
 	flagset.Usage = usageFor(flagset, "mdmctl get devices [flags]")
 	if err := flagset.Parse(args); err != nil {
 		return err
@@ -131,6 +134,15 @@ func (cmd *getCommand) getDevices(args []string) error {
 	devices, err := cmd.devicesvc.ListDevices(ctx, device.ListDevicesOption{})
 	if err != nil {
 		return err
+	}
+	if *flFilterSerial != "" {
+		filter := *flFilterSerial
+		for _, d := range devices {
+			if filter == d.SerialNumber {
+				fmt.Fprintf(out.w, "%s\t%s\t%v\t%s\n", d.UDID, d.SerialNumber, d.EnrollmentStatus, d.LastSeen)
+			}
+		}
+		return nil
 	}
 	for _, d := range devices {
 		fmt.Fprintf(out.w, "%s\t%s\t%v\t%s\n", d.UDID, d.SerialNumber, d.EnrollmentStatus, d.LastSeen)


jessepeterson:
Thanks for this! However it looks like you're filtering the device responses in the `mdmctl` tool. I think what we'd want to do is pass the input serial up to the device service and have the backend filter the devices before they're delivered. I.e. `device.ListDevicesOption` already has a `FilterSerial` array. The backend store (in this case `device.DB.List()` should probably be doing the filtering. In this way we get an efficient API transfer rather than delivering all serials to the API call just to only 'display' the filtered ones.

bdemetris:
This is probably closer to what you're interested in seeing.  Now implementing `gorilla/scheme` to grab the url params and passing them back as `opts` to `device.DB.List`

so now instead of doing this on the client side we can do something like...

`brett$ curl -X GET "https://371ccf96.ngrok.io/v1/devices?FilterSerial=SERIAL1,SERIAL2" -u "micromdm":"dev"`

and get a response like..
```
{
  "devices": [
    {
      "serial_number": "SERIAL1",
      "udid": "",
      "enrollment_status": false,
      "last_seen": "0001-01-01T00:00:00Z"
    },
    {
      "serial_number": "SERIAL2",
      "udid": "",
      "enrollment_status": false,
      "last_seen": "0001-01-01T00:00:00Z"
    }
  ]
}
```

cmd/mdmctl/get.go
```diff
@@ -128,7 +131,17 @@ func (cmd *getCommand) getDevices(args []string) error {
...
 	ctx := context.Background()
-	devices, err := cmd.devicesvc.ListDevices(ctx, device.ListDevicesOption{})
+
+	// convert string to []string in case a filter serial is defined
+	// nil serials will return all serials from the database
+	var flFilterSerialsSlice []string
+	if *flFilterSerials != "" {
+		flFilterSerialsSlice = strings.Split(*flFilterSerials, ",")
+	} else {
+		flFilterSerialsSlice = nil
```
jessepeterson: An uninitialized slice is already nil, no need to fall through here.

platform/device/builtin/db.go
```diff
@@ -54,19 +54,17 @@ func NewDB(db *bolt.DB, pubsubSvc pubsub.PublishSubscriber) (*DB, error) {
...
-	// TODO add filter/limit with ForEach
-	fmt.Println(opt.FilterSerial)
 	var devices []device.Device
 	err := db.View(func(tx *bolt.Tx) error {
 		b := tx.Bucket([]byte(DeviceBucket))
-		c := b.Cursor()
-		for k, v := c.First(); k != nil; k, v = c.Next() {
+		b.ForEach(func(k, v []byte) error {
 			var dev device.Device
 			if err := device.UnmarshalDevice(v, &dev); err != nil {
```
jessepeterson: Consider logging an error for this row instead. It could just be one bad unmarshalled record. For now though, this is probably fine.

jessepeterson: Actually, does ForEach continue on by itself?

jessepeterson:
Hey @bdemetris! Sorry this took so long to get some eyes on. Overall this looks pretty good! I added some comments. I'd like to have another look as well as check it out and give 'er a spin.

platform/device/builtin/db.go
bdemetris: nope, you're right on this. https://github.com/boltdb/bolt/blob/fd01fc79c553a8e99d512a07e8e0c63d4a3ccfc5/tx.go#L124-L126

platform/device/builtin/db.go
```diff
@@ -72,7 +74,9 @@ func (db *DB) List(opt device.ListDevicesOption) ([]device.Device, error) {
 				}
 			}
 			return nil
-		})
+		}); err != nil {
+			return err
+		}
```
jessepeterson: I'd probably just return err from the ForEach. E.g.:

```
return b.ForEach(func(k, v []byte) error {
...
})
```

platform/device/builtin/db.go
```diff
@@ -26,6 +26,7 @@ const (
 
 type DB struct {
 	*bolt.DB
+	device.Store
```
jessepeterson: Doesn't look like you're using this anywhere?

platform/device/get_devices.go
```diff
@@ -48,9 +49,13 @@ type getDevicesResponse struct {
 func (r getDevicesResponse) Failed() error { return r.Err }
 
 func decodeListDevicesRequest(ctx context.Context, r *http.Request) (interface{}, error) {
-	defer r.Body.Close()
 	var opts ListDevicesOption
-	req := getDevicesRequest{Opts: opts}
+	if err := json.NewDecoder(r.Body).Decode(&opts); err != nil {
```
jessepeterson: I would take a look at using `httputil.DecodeJSONRequest` — does a bunch of handy JSON/error handling for you. Here's an example usage: https://github.com/micromdm/micromdm/blob/e32d2bea1b56aa121abf97953ab16131d7d01e39/platform/dep/get_dep_device.go#L31-L35

jessepeterson:
This is just about good to go methinks. I had a chance to checkout your branch and give it a go. I'll approve with those couple changes! Thanks!

platform/device/get_devices.go
bdemetris: handy!

Issue #364, opened 2018-01-09
get devices failing to decode


bdemetris:
### What version of `micromdm` are you using?
`micromdm version v1.2.0-28`

### What `micromdm` command did you run?
on server
`micromdm serve -http-debug -server-url=https://server.io -api-key=dev -tls=false -http-addr=:8080`

on client
`curl -X GET https://c4d1e3ef.ngrok.io/v1/devices -u micromdm:dev`

### What did you expect to see?
expected to see a list of devices

### What did you see instead?
```
GET /v1/devices HTTP/1.1
Host: c4d1e3ef.ngrok.io
Accept: */*
Authorization: Basic bWljcm9tZG06ZGV2
User-Agent: curl/7.54.0
X-Forwarded-For: 12.17.170.22
X-Forwarded-Proto: https


---END Request---
err=EOF
---BEGIN Response---
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Type: application/json; charset=utf-8

{
  "error": "EOF"
}

---END Response---
```

### Issue
so it looks like when sending a `GET` request to `/v1/devices` the decode is failing. 

### Resolution
remove https://github.com/micromdm/micromdm/blob/175b84e6316c3b7f15c4d65be98c3714fc526ea3/platform/device/get_devices.go#L54-L56

PR #365, opened 2018-01-09
remove expectation of json from decodeListDeviceRequest

bdemetris:
testing this seems to resolve
https://github.com/micromdm/micromdm/issues/364

### DIFF
--- a/platform/device/get_devices.go
+++ b/platform/device/get_devices.go
@@ -2,7 +2,6 @@ package device
 
 import (
 	"context"
-	"encoding/json"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jessepeterson:
We should still attempt to parse the JSON, though, right? Perhaps just not return an error on parsing failure (or check for a null body)? We'll need to be able to have that data when we start filtering serials in #363.

groob:
No because GET requests should use url params not a JSON body. 

We could discuss POST vs GET but GET requests cannot have an http body. This was a bug. 

My personal preference is to implement `?page=2&per_page=20&serials=foo,bar,baz`

jessepeterson:
Well, GET requests can technically have a body. It's just merely a convention to discard it, however. But that's fine, I'm on board with URL params. There's just no code that supports decoding them at the moment. Makes me wonder if we're we doing JSON decodes in other mdmctl GET requests, too.

groob:
We had a bunch. I switched `/dep/devices` recently to a POST request because the GET had a body. 
 
The issue is that GET _could_ have a body but tools like `curl` discard it unless you explicitly pass custom flags. It's best to avoid it in a REST api service and use POST if the body is a must.

Issue #366, opened 2018-01-18
Enrollment profile rejected by iOS due to invalid topic


larsar:
### What version of `micromdm` are you using?
Tested with 1.2.0 binary and master, with the same result.

### What `micromdm` command did you run?
sudo build/linux/micromdm serve -http-debug -apns-cert ~/mdm.p12 -apns-password=redacted -server-url=https://mdm.mydomain -filerepo /home/ubuntu/pkg/ -tls-cert ~/mdm_mydomain.pem -tls-key ~/mdm_mydomain.key.pem

### What did you expect to see?
Succssfully a Mac by visiting https://mdm.mydomain/mdm/enroll
Expected that I could enroll an iOS device the same way

### What did you see instead?
When visiting the enrollment URL with Safari on iOS, I was prompted to accept the enrollment profile. After accepting, I got an error message about an "unknown error". The profile was not installed. 

The device logs indicates that the **topic** is invalid. See attached log and profile.

According to the documentation, the Topic must have a specific prefix.
https://developer.apple.com/library/content/documentation/Miscellaneous/Reference/MobileDeviceManagementProtocolRef/3-MDM_Protocol/MDM_Protocol.htm

>  Topic | String |Mandatory. The topic that MDM listens to for push notifications. The certificate that the server uses to send push notifications must have the same topic in its subject. The topic must begin with the com.apple.mgmt. prefix.

The enrollment profile contained a topic without the _com.apple.mgmt_ prefix.
```
<key>Topic</key>
<string>95PAURT649</string>
``` 

[device.log](https://github.com/micromdm/micromdm/files/1641809/device.log)
[enrollment_profile.txt](https://github.com/micromdm/micromdm/files/1641811/enrollment_profile.txt)

jessepeterson:
@larsar where did you get your push certificate?

larsar:
Got it from Apple, but I guess that 

```

larsar:
Yes, reading the [micromdm certificates](https://micromdm.io/blog/certificates/) docs now and it's clear that I'm using the wrong certificate. 

It might be an improvement if Micromdm could validate the certificate type and give an appropriate error message.

jessepeterson:
@larsar that's a great suggestion. Can you open a new issue to add that validation?

PR #373, opened 2018-01-25
Check that push certificate contains UID with required prefix

### LINKED ISSUE #367: Validation of push certificate on mdm startup
It's possible to run _micromdm serve -apns-cert cert_and_key.p12_ even though the p12 file contains a certificate that is not a push certificate. If the wrong type of certificate is used an iOS client will reject the enrollment profile.

It would be helpful for the user if the micromdm refuse to start unless the certificate is a "push certificate". 

Issue #366 shows the result of starting the mdm with a vendor certificate (aka CSR certificate) instead of a signed push certificate.

larsar:
Checks if the subject field of the push certificate contains the required prefix. Unless it does, it's probably not a push certificate. 

This issue fixes #367.
Initially I started micromdm using a "csr certificate" from Apple (ref #366). I tested the same scenario after extending the cert validation, and the server aborted as expected.

```
ts="2018/01/25 10:59:08" msg="enrollment service: get apns topic from certificate: invalid Push Topic (UserID OID) in certificate. Must start with 'com.apple.mgmt', was: 95PAURT649"
```


pkg/crypto/helpers.go
```diff
@@ -121,7 +122,12 @@ func TopicFromCert(cert *x509.Certificate) (string, error) {
 	var oidASN1UserID = asn1.ObjectIdentifier{0, 9, 2342, 19200300, 100, 1, 1}
 	for _, v := range cert.Subject.Names {
 		if v.Type.Equal(oidASN1UserID) {
-			return v.Value.(string), nil
+			var uid = v.Value.(string)
+			if strings.HasPrefix(uid, "com.apple.mgmt") {
+				return v.Value.(string), nil
+			} else {
```
groob: there's no need for an `else` branch if you have a return.

pkg/crypto/helpers.go
```diff
@@ -121,7 +122,12 @@ func TopicFromCert(cert *x509.Certificate) (string, error) {
 	var oidASN1UserID = asn1.ObjectIdentifier{0, 9, 2342, 19200300, 100, 1, 1}
 	for _, v := range cert.Subject.Names {
 		if v.Type.Equal(oidASN1UserID) {
-			return v.Value.(string), nil
+			var uid = v.Value.(string)
```
groob: this will panic if value is not a string. Unlikely but might be worth checking. 

https://golang.org/doc/effective_go.html#interface_conversions
```
str, ok := value.(string)
if ok {
    fmt.Printf("string value is: %q\n", str)
} else {
    fmt.Printf("value is not a string\n")
}
```

pkg/crypto/helpers.go
```diff
@@ -121,7 +122,12 @@ func TopicFromCert(cert *x509.Certificate) (string, error) {
...
+			if strings.HasPrefix(uid, "com.apple.mgmt") {
+				return v.Value.(string), nil
```
groob: `uid` already has the cast value.

pkg/crypto/helpers_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+func TestTopicFromCertWithWrongUidPrefix(t *testing.T) {
+	expectPushCertError("testdata/mock_push_cert_wrong_uid_prefix.pem", t)
+}
+
+func TestTopicFromCertWithoutUid(t *testing.T) {
+	expectPushCertError("testdata/mock_push_cert_no_uid.pem", t)
+}
+
+func expectPushCertError(certPath string, t *testing.T) {
```
groob: helpers should have `t.Testing` as the first argument. let's reverse this order.

pkg/crypto/helpers_test.go
```diff
@@ -0,0 +1,35 @@
...
+func TestTopicFromValidCert(t *testing.T) {
+	certificate, _ := ReadPEMCertificateFile("testdata/mock_push_cert.pem")
+	pushTopic, err := TopicFromCert(certificate)
+
+	if err != nil {
+		t.Errorf("fail %s", err)
+	}
+
+	if have, want := pushTopic, "com.apple.mgmt.External.18a16429-886b-41f1-9c30-2bd04ae4fc37"; have != want {
+		t.Errorf("have %d, want %d", have, want)
```
groob: %d is for integers. This should be a `%s`

pkg/crypto/helpers_test.go
```diff
@@ -0,0 +1,35 @@
+package crypto
+
+import (
+	"testing"
+)
+
+func TestTopicFromValidCert(t *testing.T) {
+	certificate, _ := ReadPEMCertificateFile("testdata/mock_push_cert.pem")
+	pushTopic, err := TopicFromCert(certificate)
+
```
groob: remove the extra line. `if err != nil` should be grouped with the TopicFromCert call.

pkg/crypto/helpers_test.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		t.Errorf("fail %s", err)
```
groob: You can use `t.Fatalf` here to stop the test execution if an error was returned. 
In case of `pushTopic` it doesn't matter too much because it's a string with a default value of `""`, but if you had a function which returned a pointer, the default value would be nil, likely causing a panic if the test continued. 

https://golang.org/pkg/testing/#T.Fatalf

pkg/crypto/helpers_test.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	if have, want := pushTopic, "com.apple.mgmt.External.18a16429-886b-41f1-9c30-2bd04ae4fc37"; have != want {
+		t.Errorf("have %d, want %d", have, want)
+	}
+}
+
+func TestTopicFromCertWithWrongUidPrefix(t *testing.T) {
```
groob: IMO these tests are a bit too short. A common go pattern is to write a [table test](https://github.com/golang/go/wiki/TableDrivenTests).

[Subtests](https://blog.golang.org/subtests) also help a lot when working with a table test. 

And here's an example in the wild. https://github.com/oklog/oklog/blob/7fa04d3c21f8fedb54940c069c02e450c7cc8eb3/pkg/store/overlap_test.go

larsar:
Thanks for the thorough review, @groob. It's great way to learn! :)

pkg/crypto/helpers.go
larsar: I'm not sure I understand you correctly on this one. If no UID is found, the function returns an error. If the UID is present, then the code will check if it has the required prefix or not. In either case it should break the loop, which the return does, but in one case it should return the UID value, in the other case it should return an error. 

If I remove the else block the method will return the error message at the bottom of the method instead a more explanatory error message defined in the else block.

pkg/crypto/helpers_test.go
larsar: Like this?
```
	if pushTopic, err := TopicFromCert(certificate); err != nil {
		t.Errorf("fail %s", err)
	} else {
		if have, want := pushTopic, "com.apple.mgmt.External.18a16429-886b-41f1-9c30-2bd04ae4fc37"; have != want {
			t.Errorf("have %d, want %d", have, want)
		}
	}
```

Not sure I like to nest the next test inside an else block, but the pushTopic variable will be scoped to the if block if do it like this.

Sorry for my n00b Go knowledge. I watched a Go crash course on high speed before starting coding. :)

pkg/crypto/helpers_test.go
larsar: Doh.... Last minute refactoring with some copy/paste. :)

pkg/crypto/helpers.go
larsar: Wilco

pkg/crypto/helpers_test.go
larsar: I agree. Fatal it is. :)

pkg/crypto/helpers_test.go
larsar: Wilco

pkg/crypto/helpers_test.go
larsar: I'll check out the table test patterns. I agree that a whole function for "one liner tests" is not ideal.

pkg/crypto/helpers_test.go
groob: Sorry what I meant by this comment is to remove the newline between TopicFromCert and the `if err != nil` (line 10)

The actual code is fine.

pkg/crypto/helpers.go
groob: > If no UID is found, the function returns an error

and function execution stops because the error was returned. The behavior is the same wether `else` is present or not. 
All I'm advocating for is to remove the extra branch and keep the code from within the `else` statement at a different indentation level. 

This comment would also come from `golint` I believe. 

Take this example: 
```
func get(i int) int {
    if f := i * 2; f == 6 {
        return f
    } else {
        return f * 3
    }
}
``` 

The code is correct, but the recommended go style is to write it like this:

```
func get(i int) int {
    f := i * 2
    if f == 6 {
        return f
    }
    return f * 3
}
```

larsar:
Squashed and pushed now. :)

Issue #293, opened 2017-11-01
Support using user-targeted MDM


jessepeterson:
#274 added user-targeted MDM support code/API. To add support to be able to use user-targeted MDM with built-in tools.

jessepeterson:
Need to build support for user-targeted MDM for OD authenticated users. I.e. we need to be storing  UserAuthenticate messages and keeping track of user-based Tokens. Also to support verification/authentication of Digest password that's handed to the MDM server for per-user auth.

jessepeterson:
Just a further note here. This is issue is for the support for network account MDM target which is authenticated differently than the single "local user" MDM account notification (i.e. they follow the separate `UserAuthenticate` checkin protocol).

PR #379, opened 2018-02-03
Reject UserAuthentication attempts per MDM spec

jessepeterson:
I could have sworn we were doing this already, but I think I got confused with [commandment](https://github.com/cmdmnt/commandment/blob/9258be54b9ba9c0595768da02fa6ee8fb6054163/commandment/mdm/app.py#L155-L157) where it was implemented it. :)

In any case this will reject UserAuthentication attempts with the special 410 (Gone) HTTP status code per the MDM spec. This is for two reasons. One is when I recall in my testing that when you reject the UserAuthenticate message you don't get [the prompt for management settings](https://derflounder.wordpress.com/2013/11/13/bypassing-the-mavericks-managed-preferences-login-check/). The other is that we probably shouldn't be sending the wrong response. We appear to be sending an a 401 response for unknown Checkin request MessageTypes which I'm not sure how mdmclient handles for e.g. the UserAuthenticate.

### DIFF
--- a/mdm/checkin/endpoint.go
+++ b/mdm/checkin/endpoint.go
@@ -27,14 +27,12 @@ func MakeCheckinEndpoint(svc Service) endpoint.Endpoint {
 		case "CheckOut":
 			err = svc.CheckOut(ctx, req.CheckinCommand)
 		case "UserAuthenticate":
+			// TODO: to support per-user MDM. See #293
 			err = &rejectUserAuth{}
 		default:
 			err = errInvalidMessageType
 		}
-		if err != nil {
-			return checkinResponse{Err: err}, nil
-		}
-		return checkinResponse{}, nil
+		return checkinResponse{Err: err}, nil
 	}
 }
 
@@ -57,8 +55,6 @@ func (e *rejectUserAuth) Error() string {
 func (e *rejectUserAuth) UserAuthReject() bool {
 	return true
 }
-
-func (e *rejectUserAuth) error() error { return e }
 
 func isRejectedUserAuth(err error) bool {
 	type rejectUserAuthError interface {
--- a/mdm/checkin/endpoint.go
+++ b/mdm/checkin/endpoint.go
@@ -2,10 +2,10 @@ package checkin
 
 import (
 	"context"
-	"errors"
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/micromdm/mdm"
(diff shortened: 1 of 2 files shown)


mdm/checkin/endpoint.go
```diff
@@ -26,8 +26,10 @@ func MakeCheckinEndpoint(svc Service) endpoint.Endpoint {
 			err = svc.TokenUpdate(ctx, req.CheckinCommand)
 		case "CheckOut":
 			err = svc.CheckOut(ctx, req.CheckinCommand)
+		case "UserAuthenticate":
```
groob: As implemented this change would prevent the admin user account as installed by the blueprint. No?

jessepeterson: I don't think so, but we should test to be sure. The user MDM checkin (UserAuthenticate) happens post user-login I think. In any case responding with a 401 can't be any better. :)

mdm/checkin/endpoint.go
```diff
@@ -45,3 +47,25 @@ type checkinResponse struct {
...
+
+func (e *rejectUserAuth) Error() string {
+	return "reject user auth"
+}
+
+func (e *rejectUserAuth) UserAuthReject() bool {
+	return true
+}
+
+func (e *rejectUserAuth) error() error { return e }
```
groob: this method is unnecessary. `Error() string` is the one that implements the interface.

jessepeterson: Doesn't that need to be there because that's how it implements the `errorer` interface? Which is what in turn the `func encodeResponse` checks for?

groob: You're right. but that method doesn't go on the error. It is a method on the `checkinResponse`

But that brings up a good point. You should not be returning the error directly, but instead returning the checkinResponse struct, with the Err field set to your error. 

```
return checkinResponse{Err: errRejectUserAuth}, nil
```

The encode response will see that the checkinResponse is an `erroer` and pass it to the error encoding function.

jessepeterson: This doesn't return the error directly. All errors there are wrapped in a `checkinResponse`. However I do see that `errorer` is on the response object and not on the error itself — its superfluous here. I'll update.

jessepeterson:
See #293 for adding support in the future.

groob:
this is good to ship

PR #387, opened 2018-03-15
updated added another 0 to the  value

danfalzon-sohonet:


platform/command/transport_http.go
```diff
@@ -47,7 +47,7 @@ func EncodeError(ctx context.Context, err error, w http.ResponseWriter) {
 
 func decodeRequest(ctx context.Context, r *http.Request) (interface{}, error) {
 	var req newCommandRequest
-	err := json.NewDecoder(io.LimitReader(r.Body, 100000)).Decode(&req)
+	err := json.NewDecoder(io.LimitReader(r.Body, 1000000)).Decode(&req)
 	return req, err
```
groob: could you also change the return from 
```
return req, err
```

to 

```
return req, errors.Wrap(err, "decoding command request")
```

danfalzon-sohonet: Just updated

groob:
@danfalzon-sohonet this now requires the import (top of the file) to have `github.com/pkg/errors`, which is the library that handles the errors wrap. Tests should pass after that.

Issue #320, opened 2017-11-13
Proposal: limited enrolling slots


Beanow:
To restrict how a device might be enrolled, using a concept of slots.
(#310 / #81)

1. Call the API commands to create a new unique slot.
    Produces an enroll link for this slot `/mdm/enroll/<Slot ID>`.
    A random SCEP challenge is created and included.
    Slot is stored in the DB with the ID, the **served** state.
1. The profile served on that link will contain the same Slot ID on `/mdm/checkin/<Slot ID>`.
    In this mode, requests without a known Slot ID will be rejected.
    If accepted the enrolling proceeds as normal.
    Slot is updated in the DB to have the UDID and the **consumed** state.

An additional API endpoint and mdmctl command will allow itterating all the slots.

This will allow an implementer to have an "invite-only" scheme where your application can decide what preconditions will warrant a slot to be created.

And after the fact this slot can be associated with a device once enrolling has completed. Allowing the application to tie this device into a broader application context.

groob:
That's an interesting idea. We definitely need to address who is allowed to enroll and have discussed several possible ways of doing it, but never this. 

Let's wait and see what input comes from other users/devs.

Beanow:
Although I'm guessing at their implementation details, it's a system that https://simplemdm.com/ applies to associate a particular device with a user.

groob:
@jessepeterson @bruienne @np5 I'd be curious as to thoughts about this proposal.

jessepeterson:
**tl;dr:** yes; we've talked about systems like this and their benefits. as always devil's in the details. 😄 

This is essentially the idea with OTA enrollments and why that stub code exists. Your Slot ID is essentially what we've been calling a token all along. How it initially gets "issued" would be a matter up to the admin/implementor.

With this specific proposal I'm just not sure why the slot is passed around so often in the URL endpoints? On each request to CheckInURL and Command endpoints you are authenticated against the client certificate. If the client certificate is generated from the SCEP request the only real place this "slot" (or token) is needed is in the SCEP challenge field of the enrollment profile (i.e. there's an inherent association between the device cert and the SCEP challenge). No need to pass it around on the endpoints. Of course if you don't have control of the SCEP server you may not be able to get that association (but perhaps we proxy SCEP in this case to capture issued certificates). But if you don't have access to the SCEP server "unique" challenges may not be possible anyway.

[...]

PR #389, opened 2018-03-23
OTA Enrollment with an enroll invite id and device id webhook

larsar:
Currently, this is only a prototype implementation. This pull request is mainly to share the code, not to actually merge it. 

**Objective**
I need to start the enrollment process from an external system. The external system generates an enrollment "challenge" and needs to link it to the device id after enrollment.

**Prototype implementation**
* Read the enrollment challenge from URL parameter (/ota/enroll?challenge=12345). 
* Use the supplied challenge for the first profile given to the device
* The device then returns both the challenge and device id as a part of the enrollment request.
* The device id and invite id is then sent via a webhook to the external service as a json:
```
{"challenge":"1234","event":"ota_phase1","udid":"7c31b3a19a0aa12dxxxxxxxxxxx7531e1c1fd7e"}
``` 

* Web hook triggered after enrollment has completed:
```
{"event":"enroll","udid":""7c31b3a19a0aa12dxxxxxxxxxxx7531e1c1fd7e"}
```

* After checkout:
```
* {"event":"checkout","udid":""7c31b3a19a0aa12dxxxxxxxxxxx7531e1c1fd7e"}
```

This code example might be relevant for the discussions in #320

### DIFF
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -82,6 +82,8 @@ const homePage = `<!doctype html>
 <body>
 	<h3>Welcome to MicroMDM!</h3>
 	<p><a href="mdm/enroll">Enroll a device</a></p>
+	<p><a href="ota/enroll">Enroll a device OTA</a></p>
+
 </body>
 </html>
 `
@@ -104,6 +106,7 @@ func serve(args []string) error {
 		flDepSim            = flagset.String("depsim", "", "use depsim URL")
 		flExamples          = flagset.Bool("examples", false, "prints some example usage")
 		flCommandWebhookURL = flagset.String("command-webhook-url", "", "URL to send command responses as raw plists.")
+		flEnrollWebhookURL  = flagset.String("enroll-webhook-url", "", "URL to send invitation id and UDID during enrollment.")
 	)
 	flagset.Usage = usageFor(flagset, "micromdm serve [flags]")
 	if err := flagset.Parse(args); err != nil {
@@ -139,6 +142,7 @@ func serve(args []string) error {
 		depsim:              *flDepSim,
 		tlsCertPath:         *flTLSCert,
 		CommandWebhookURL:   *flCommandWebhookURL,
+		EnrollWebhookURL:    *flEnrollWebhookURL,
 
 		webhooksHTTPClient: &http.Client{Timeout: time.Second * 30},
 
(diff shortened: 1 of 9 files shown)


larsar:
I have updated the pull request, @groob. The mechanism works fine, but I have do some more cleanup before I'm done. 

Currently this only makes it possible to link external enroll request with device using the OTA enrollment. I believe we should provide the same mechanism for the regular MDM enrollment, but then the challenge has to be sent from the MDM to the SCEP service (I think) and the challenge + udid must later be returned.

larsar:
Closing this one since I'm probably not going to continue down the "OTA enrollment" road for now.

jessepeterson:
@larsar: any particular reason why you're pursuing less secure enrollment delivery mechanisms?

larsar:
I don't see why the OTA enrollment is any more secure than the MDM enrollment: A HTTP request over a TLS connection results in a MDM profile being installed. The OTA enrollment has an extra roundtrip that might be used for an additional authentication step, but currently MicroMDM does not utilize this.

The main issue is that the OTA enrollment sometimes fails. iOS refuses to install a profile, and the enrollment process fails. It usually works when clicking the install button one more time. Our iOS app was rejected by app review because of this issue, and since we're not able to reproduce the problem (it seems to occur randomly) we decided to use the simpler MDM enrollment process. The only practical consequence is that we need to add an identifier in the checkin URL instead of using an initial enrollment profile with a user specific challenge. 

Pull request #399 contains our current solution. I might get back to the OTA enrollment, and reopen this pull request, when I have more time to debug the error that randomly occurs.

jessepeterson:
Well the main thing that protects DEP and OTA enrollment is the Apple device-only client-signed PKCS7 request. In other words _only_ Apple devices will be able to to reach those end-points (assuming the MDM server validates that certificate [chain], which MicroMDM does). Secondary to that you get information, like you do with DEP, on the device itself which you can base enrollment decisions on (like serial number, IMEI, UDID, etc.).The other thing, the two-step SCEP process is of arguable security, but considering TLS and cert. pinning I'd argue it doesn't add too much, as you said. I wish the above were possible without the two-step SCEP, honestly — I think its a relic from being able to use OTA enrollment over non-HTTPS endpoints (and MDM requires HTTPS, etc.). It's true MicroMDM doesn't have any 3rd-party authenticated enrollment (other than the above-mentioned device certificate CA checking for DEP & OTA).

I can say that OTA enrollment on a _Mac_ seems to be very reliable. We have a large fleet enrolled using that method with pretty much zero issues. However it's interesting you're getting enrollment failures sometimes on iOS. I haven't tested much with that and indeed needs further testing. I'd be curious if you had any logs or any information to debug that further! Too bad it fails randomly! Did you ever notice it failed on the same device? What was frequency of failures?

Issue #388, opened 2018-03-22
mdmctl should display a sensible error message when API key is wrong or not set by server


larsar:
If the server is started without specifying an API key, or that the configuration file for mdmdctl contains the wrong key, a misleading error message is displayed. The error message should explain that it's an authentication error, not a parsing error. 

### What version of `micromdm` are you using?
micromdm version v1.2.0-35-g185aa41-dirty

### What `micromdm` command did you run?
mdmctl get devices

### What did you expect to see?
micromdm> mdmctl get devices
UDID SerialNumber EnrollmentStatus LastSeen
7c31b3e19d0aa32de8b97fcb69cb753ae1c1fd7e F12LM4DYXFGC true 2018-03-22 17:18:18.542328 +0000 UTC

### What did you see instead?
```
micromdm> mdmctl get devices
UDID SerialNumber EnrollmentStatus LastSeen
expected JSON formatted error, got Content-Type text/plain; charset=utf-8
```

PR #396, opened 2018-04-15
must start server with api-key

bdemetris:
#388

### DIFF
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -118,9 +118,6 @@ func serve(args []string) error {
 	if *flServerURL == "" {
 		return errors.New("must supply -server-url")
 	}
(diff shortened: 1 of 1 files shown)


jessepeterson:
I'm against this change fwiw. The server should have read-ish-only mode.

larsar:
What about printing a warning message if api-key is not specified?

bdemetris:
while poking at the linked issue i didn't see value in starting the server without an api key since there are very few routes that aren't authenticated.  on second thought i suppose that is a bit heavy handed, and there seems to be thought already invested on this matter since there is currently a catch for empty api key.

https://github.com/micromdm/micromdm/blob/master/cmd/micromdm/serve.go#L326

a warn seems reasonable

PR #399, opened 2018-04-22
Use externally set as parameter in checkin URL and send JSON webhook callback

larsar:
If an id parameter is specified in the enroll URL /mdm/enroll?id=1234, then
use this in the checkin url /mdm/checkin?id=1234 in the enrollment profile.

If checkin webhook is specified, the id and device id (UDID) will be encoded
as JSON and sent when the device check in and out.

The JSON webhook is generic and can be used for other purposes than checkin
events.

### DIFF
--- a/mdm/enroll/service.go
+++ b/mdm/enroll/service.go
@@ -23,7 +23,7 @@ const (
 )
 
 type Service interface {
-	Enroll(ctx context.Context) (profile.Mobileconfig, error)
+	Enroll(ctx context.Context, id string) (profile.Mobileconfig, error)
 	OTAEnroll(ctx context.Context) (profile.Mobileconfig, error)
 	OTAPhase2(ctx context.Context) (profile.Mobileconfig, error)
 	OTAPhase3(ctx context.Context) (profile.Mobileconfig, error)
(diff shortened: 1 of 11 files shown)


mdm/enroll/service.go
```diff
@@ -167,13 +167,21 @@ func (svc *service) findOrMakeMobileconfig(id string, f interface{}) (profile.Mo
 	return p.Mobileconfig, nil
 }
 
-func (svc *service) Enroll(ctx context.Context) (profile.Mobileconfig, error) {
-	return svc.findOrMakeMobileconfig(EnrollmentProfileId, svc.MakeEnrollmentProfile)
+func (svc *service) Enroll(ctx context.Context, id string) (profile.Mobileconfig, error) {
+	if id == "" {
+		return svc.findOrMakeMobileconfig(EnrollmentProfileId, svc.MakeEnrollmentProfile)
```
larsar: This does not work. When MakeEnrollmentProfile takes an id argument, the findOrMakeMobileconfig method fails. I know what the problem is, but I'm not sure what the best way to add the id to the checkin URL with minimal code changes. Do you have any suggestions, @groob ?

jessepeterson:
I've only had the briefest of looks at this but would you consider implementing different `checkin` and `enroll` services to support this use case? Something like `idcheckin` and `idenroll` services (which are then passed into the URL handlers conditionally). Just thinking of ways we can decouple the ID logic from the core functionality.

larsar:
That seems like a good idea, @jessepeterson.

groob:
@larsar I ended up redoing both webhooks and the `/mdm/connect`, `mdm/checkin` routes to forward any URL params from those route to the webhook. 

I also update the webhook worker to support checkins https://github.com/micromdm/micromdm/pull/424/files 

Would love to know what you think.

knightsc:
@larsar Have you had a chance to use the latest webhook changes? Does the current implementation do everything you need it to? Do you think we should keep this PR open?

PR #401, opened 2018-05-01
Implement mdmcert.download support in mdmctl

jessepeterson:
Death to `certtool`! 😄

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -21,7 +21,7 @@
   branch = "master"
   name = "github.com/fullsailor/pkcs7"
   packages = ["."]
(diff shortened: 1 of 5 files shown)


jessepeterson:
Of note pkcs7 needed to updated due to this bug: fullsailor/pkcs7#31 (go 1.10 issue) which was fixed.

jessepeterson:
Also — last minute thought — this really ought to go under the `mdmcert` sub-command and not be its own top-level command.

jessepeterson:
Updated wiki docs as well: https://github.com/micromdm/micromdm/wiki/mdmcert.download

Issue #178, opened 2017-05-28
Appropriately handle OpType in DEP sync


jessepeterson:
Currently we unconditionally add new devices to the device DB but we should better handle the specific OpType returned by the DEP sync API. Also noted as [a TODO in device/db.go](https://github.com/micromdm/micromdm/blob/8b36409f2061a34ffcd39cacb662439803388699/device/db.go#L304)

PR #405, opened 2018-05-07
DEP auto-assignment (v2)

### LINKED ISSUE #227: add DEP profile assignment on sync
manually adding a list of serials to a DEP profile is not a good solution for production. 

Add the ability to auto-assign devices a DEP profile template if a new device shows up after DEP sync. 
The rules to do this should support a device type predicate (mac, ios, tvos)

jessepeterson:
The rework of #326 after the repository layout change with some reworking. Will resolve #227 & #178.


jessepeterson:
In-progress notes:

Suggested changes implemented in this diff from previous PR comments/code rview. Also refactored the Gokit interfaces to match the newer service layout (i.e. service/API-based source files). Right now there's a bit more work to do:

- [ ] Ability to set auto-assign profile upon profile _creation_ (simple change)
- [ ] Implement getter and deleter for AutoAssigner filters in `mdmctl` (setting already exists)

The ability to re-assign the existing enrolled machines will be another PR. This probably needs some discussion/RFC on decoupling of the device and depsync services (e.g. do we send all devices over the pubsub interface; another internal-only interface, etc?)

jessepeterson:
Just want to add a note about what UX looks like here. In general nothing more than adding a single switch to a `mdmctl apply dep-profiles` command should be needed. For example what used to be this:

```
$ mdmctl apply dep-profiles -f path/to/dep-profile.json
Defined DEP Profile with UUID 3462FDA3E14D4870
```

Is now:

```
$ mdmctl apply dep-profiles -f dep-profile.json -filter='*'
Defined DEP Profile with UUID 3462FDA3E14D4870
Saved auto-assign filter '*' for this DEP profile
```

From then on the auto-assigner will apply that profile UUID to all serials matching the filter (at this time, all serials match the `*` filter). You can also manage the filters using the `mdmctl apply dep-autoassigner` et. al. commands but doing it all at once while setting the DEP profile is the easiest.

dep/depsync/apply_autoassigner.go
```diff
@@ -0,0 +1,57 @@
+package depsync
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/go-kit/kit/endpoint"
+
+	"github.com/micromdm/micromdm/pkg/httputil"
+	// 	"github.com/micromdm/micromdm/dep/depsync"
```
groob: remove commented line

dep/depsync/server.go
```diff
@@ -0,0 +1,67 @@
...
+	))
+
+	r.Methods("POST").Path("/v1/dep/autoassigners").Handler(httptransport.NewServer(
+		e.ApplyAutoAssignerEndpoint,
+		decodeApplyAutoAssignerRequest,
+		httputil.EncodeJSONResponse,
+		options...,
+	))
+
+	// r.Methods("GET").Path("/v1/dep/account").Handler(httptransport.NewServer(
```
groob: why are these commented out?

jessepeterson: No reason, they shouldn't be there. They're vestiges from the service (dep) I copied them from.

jessepeterson:
Added Wiki docs for this feature: https://github.com/micromdm/micromdm/wiki/DEP-auto-assignment

Issue #226, opened 2017-09-05
mdmctl: Add "first seen" to list of devices 


groob:
We output `LastSeen` as a column when we run `mdmctl get devices`. 
```
~ ❯❯❯ mdmctl get devices  
UDID                                  SerialNumber  EnrollmentStatus  LastSeen
```

This information is very useful for enrolled hosts, but makes it hard to sort by new DEP additions (which have not been enrolled yet). Printing FirstSeen would display newly added DEP devices.

groob:
This is where we print the current fields 
https://github.com/micromdm/micromdm/blob/4f86fe5679518592c6439a1c5f33a7d76215662e/cmd/mdmctl/get.go#L128

jessepeterson:
Note this issue is actually consolidation for a few other items:

* Make code variable names consistent with functions (e.g. `LastCheckin` -> `LastSeen`)
  * This implies then making a new `LastCheckin` variable to track proper MDM checkins
* Per this issue's title add a `FirstSeen` flag
* Per a note in #251 add a `LastEnrollment` flag

PR #409, opened 2018-05-08
Rename LastCheckin to LastSeen

jessepeterson:
It's a small thing, variable naming, but I've long thought conflating the term Checkin doesn't help readability. One more thing completed for #226 (remaining thing to add the actual first seen field).


jessepeterson:
Note I did not change the protobuf definition itself as people will have data corresponding still in active databases. I _think_ you can freely change the names/symbols as long as the data/field number stays the same, but not 100% there.

groob:
Yes renaming proto fields should be safe. 

I do think it's important to differentiate between the two though. LastSeen and LastCheckin have different meanings and different behaviors. What is the original field actually tracking `/mdm/checkin` or `/mdm/connect` times?

Issue #371, opened 2018-01-23
do not allow `mdmctl config set` without args


groob:
currently it creates an empty entry

```
{
 "active": "",
 "servers": {
   "": {
     "api_token": "",
     "server_url": "",
     "skip_verify": false
   },
```

jessepeterson:
I'd argue for allowing an empty name. Otherwise agree!

scottknight:
@groob Should this get closed since #421 was merged?

jessepeterson:
@scottknight FYI you can have PRs auto-close issues by including a "Fixed #<issuenum>": https://help.github.com/articles/closing-issues-using-keywords/

:)

PR #421, opened 2018-05-21
Do not allow `mdmctl config set` without args

knightsc:
Fixes issue in https://github.com/micromdm/micromdm/issues/371

I made the `-server-url` and `-api-token` flags required parameters. Additionally if set is called with no arguments at all the help is printed.

I also added a fix to make sure the switch command only allows you to switch to a server config that exists

### DIFF
--- a/cmd/mdmctl/config.go
+++ b/cmd/mdmctl/config.go
@@ -138,11 +138,18 @@ func setCmd(args []string) error {
 		return err
 	}
 
+	if flagset.NFlag() == 0 {
+		flagset.Usage()
+		os.Exit(1)
+	}
+
 	cfg := new(ServerConfig)
 
-	if *flToken != "" {
-		cfg.APIToken = *flToken
+	validatedToken, err := validateAPIToken(*flToken)
+	if err != nil {
+		return err
 	}
+	cfg.APIToken = validatedToken
 
 	validatedURL, err := validateServerURL(*flServerURL)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


bdemetris:
see outcome of https://github.com/micromdm/micromdm/pull/396.  it might help you on the direction you are going with validating the api token.

knightsc:
@bdemetris Were you thinking that APIToken shouldn't be required? I really based this on the comments in the issue https://github.com/micromdm/micromdm/issues/371. I made `validateAPIToken` a method more so to follow the pattern already being used to validate the server url more so than the fact that we're really validating something about the `APIToken`. I'm not sure if `mdmctl` has logging in the same way the server does?

bdemetris:
haha, i see what you're saying.  fwiw im personally in favor of your proposed change, since i dont see a lot of value of starting the server without an api key - but its possible.  im also not an authority on RPC - but as i understand it the client is an extension of the server - so if the server can start without a token i would think the client should also be able to start without one.  this is more of a topic for the owners to address - i was just trying to provide you with some context 😄

jessepeterson:
To be clear this is just for `mdmctl`, yeah? Seems reasonable to require an API-key there. You can't talk to a MicroMDM server that has an an empty API key — the server treats that as disabling the API endpoints.

cmd/mdmctl/config.go
```diff
@@ -182,20 +189,30 @@ func migrateCmd(args []string) error {
...
+	if err != nil {
+		return "", err
+	}
+	u.Path = "/"
+	serverURL = u.String()
+
 	return serverURL, nil
+}
+
+func validateAPIToken(apiToken string) (string, error) {
```
groob: IMO this function is unnecessary, just check the string directly before assigning.

knightsc: Updated in fixup bfbe930

Issue #402, opened 2018-05-02
Mismatched UDIDs and serial numbers cause issues?


jessepeterson:
Using current code I think there _may_ be issues with mismatched UDIDs and serial numbers. My workflow was basically spin up an entirely different VM using the same DEP serial number. This caused  two notable issue a) MicroMDM assuming this was a new device (which, from a UDID standpoint is perhaps true) and b) command queuing/delivery appeared to fail for this newly registered 'device' using the same serial number. I suspect the UDID and/or device Indexing boltdb buckets may be partly to blame here.

I haven't had time to delve in more, just wanted to note to come back to later. This may be causing issues with people trying a new/different VM in their testing workflows using the same serial number and running into problems.

jessepeterson:
Maybe #321 related?

knightsc:
I've managed to get in a related weird state through using a VM and changing things. I have two entries in a test micromdm instance that both have the same UDID but different serial numbers. With real hardware I wouldn't think this would really happen but it does seem like it might be useful for the server to consider certain things unique and not allow re-registering.

PR #422, opened 2018-05-26
Prevent multiple device records with the same UDID

knightsc:
The error I was seeing was if I changed the serial number in a VM and then re-enrolled the server would create a second record for the same UDID. I recreated this condition using the ngrok set ups and then saving an example of a `checkin` request. I think changed the serial number and replayed and each time I changed and replayed I got a new device record.

Looking at the device db it looks like we weren't using the device we looked up in the `DeviceByUDID` call.

After apply this patch and retrying my testing above I instead see the serial number updated on the existing record vs a new record with the same udid.

This might be related to https://github.com/micromdm/micromdm/issues/402 but that case is slightly different. That was the udid changing but the serial number staying the same and push not working properly.

### DIFF
--- a/platform/device/builtin/db.go
+++ b/platform/device/builtin/db.go
@@ -218,7 +218,7 @@ func (db *DB) pollCheckin(pubsubSvc pubsub.PublishSubscriber) error {
 					fmt.Println(err) // some other issue is going on
 					continue
 				}
-				_, err = db.DeviceByUDID(ev.Command.UDID)
+				byUDID, err := db.DeviceByUDID(ev.Command.UDID)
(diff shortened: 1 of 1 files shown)


knightsc:
I've been trying to think of a good place to possibly add a unit test for this error. Maybe it makes sense to add a new test case to the `checkin_test.go` file for this. The alternative that I was thinking was somehow to unit test the logic in the `device.DB` event handlers.

groob:
If you would like to add a unit test for this it should be added in `builtin/db_test.go`. 
The checkin tests do not test enrollment, they just test that the message is decoded. 

At some point it would be good to add some end to end tests but for a unit test, the device db component should be tested independently.

PR #424, opened 2018-05-27
Add a more complete webhook worker.

groob:
Adds a webhook worker which is scaffolded to support multiple event types.
Supports Checkin And Acknowledge events from the start.


groob:
@jessepeterson that's fair. 
I used Acknowledge because that's what I ended up calling the method but we can rename it. 
I have a few other PRs pending which assume the new name, so I'm going to do a search and replace after I merge all of them.

PR #428, opened 2018-05-30
UDID/device certificate matching

### LINKED ISSUE #358: Match UDID to authentication
Right now we trust that a device gives us its correct UDID in the command (connect) and checkin endpoints. However in this way a compromised device cert and/or rogue device (however unlikely) is able to read any given device's command queue by simply forging another UDID.

To consider implementing a check in those endpoints (or some middleware, etc.) that checks that a provided UDID matches the authentication (client certificate/cert. signature) given to us in the request.

jessepeterson:
Overview:

* When verifying the MDM signature, returns the device certificate now
* Includes the device certificate in the Checkin event
* Reads & writes this device cert to the device database (could be just a hash per below TODO)
* Implements Middleware to do actual UDID/cert matching against device database

Resolves #358.

TODO:

- [ ] Extend certificate checking to Acknowledge service middleware method.
- [ ] Merge with refactored device service (#427)
- [ ] Return something better than a 500 error. Perhaps a 401?
- [ ] Consider only saving a (SHA256?) hash of the certificate instead of the whole DER. Or maybe the same attributes used to determine OCSP status.

Looking for feedback!


mdm/checkin_event.go
```diff
@@ -10,11 +11,12 @@ import (
...
-	Time    time.Time
-	Command CheckinCommand
-	Params  map[string]string
-	Raw     []byte
+	ID         string
+	Time       time.Time
+	Command    CheckinCommand
+	Params     map[string]string
+	Raw        []byte
+	DeviceCert *x509.Certificate
```
groob: can this be `[]byte`? It will be easier if it doesn't have to be decoded/parsed every time.

mdm/server.go
```diff
@@ -48,29 +49,36 @@ func RegisterHTTPHandlers(r *mux.Router, e Endpoints, verifier SignatureVerifier
...
+func (d *requestDecoder) readBody(r *http.Request) ([]byte, *x509.Certificate, error) {
 	defer r.Body.Close()
 
 	body, err := ioutil.ReadAll(r.Body)
 	if err != nil {
-		return nil, errors.Wrap(err, "reading MDM Response HTTP Body")
+		return nil, nil, errors.Wrap(err, "reading MDM Response HTTP Body")
 	}
 
+	// TODO: If we ever use Go client cert auth we can use
```
groob: One thing to be cautious of here is that if we use TLS auth, we cannot reject via the handler. It has to be done at the TLS layer, before the handler has a chance to work with the request.

mdm/mw/udidauth.go
```diff
@@ -0,0 +1,86 @@
+package mw
```
groob: There's already a few middlewares around the ack/checkin endpoints. For example, `platform/remove` is a middleware that checks wether the device is blacklisted before allowing the request. 

I'd prefer if we didn't have a generic package name like `mw` for something more specific. Maybe `package mdmauth`?

mdm/mw/udidauth.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (mw *udidAuthMiddleware) Acknowledge(ctx context.Context, req mdm.AcknowledgeEvent) ([]byte, error) {
+	// fmt.Println("CERT<->UUID CHECKING HERE", req.Command.UDID, req.DeviceCert)
+	return mw.next.Acknowledge(ctx, req)
+}
+
+func (mw *udidAuthMiddleware) Checkin(ctx context.Context, req mdm.CheckinEvent) error {
+	switch req.Command.MessageType {
+	case "Authenticate":
```
groob: Depending on the pubsub implementation, there might be some latency between a message being published and processed. This approach here seems problematic because it assumes it will be possible for Authenticate to be processed just in time for the TokenUpdate.

platform/device/internal/deviceproto/device.proto
```diff
@@ -33,5 +33,5 @@ message Device {
     string dep_profile_assigned_by =27;
     int64 last_seen =28;
     bytes last_query_response =29;
-
+    bytes device_cert = 30;
```
groob: IMO the device cert lives with the CA not the device record.

mdm/mw/udidauth.go
```diff
@@ -0,0 +1,86 @@
...
+		// upgrade any devices that don't have a certificate to have one
+		devByUUID.DeviceCert = cert
+		err = mw.store.Save(devByUUID)
+		if err != nil {
+			return false, err
+		}
+		fmt.Println("SAVED NEW!")
+		return true, nil
+	}
+	if bytes.Compare(cert.Raw, devByUUID.DeviceCert.Raw) == 0 {
```
groob: `bytes.Compare` is not suitable for security checks because it's sensitive to timing attacks. https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare is a good substitute.

mdm/mw/udidauth.go
```diff
@@ -0,0 +1,86 @@
...
+	return func(next mdm.Service) mdm.Service {
+		return &udidAuthMiddleware{
+			store: store,
+			next:  next,
+		}
+	}
+}
+
+type udidAuthMiddleware struct {
+	store *devicebuiltin.DB
```
groob: import or define an interface, but don't import the concrete implementation. Trying to phase these types of imports out to allow for a Postgres Implementation :)

This middleware only uses the `DeviceByUDID` and `Save` methods. IMO the best approach would be to define a local interface in this package with just those two methods.

groob:
@jessepeterson My one concern/discussion point for this change is the reliance on the async approach (pubsub, which can have significant network latency) for implementing the certificate verification. IMO a better approach for the middleware would be to define an RPC endpoint which can verify the certificate and the middleware can call the endpoint / get back a response with a reasonable timeout. 

But this kind of mechanism already exists in micromdm/scep and there are several verification mechanisms supported by the CA. Any thoughts on why implementing verification here is desirable instead of doing that work in the CA?

jessepeterson:
@groob I absolutely agree there is race condition potential here. The same with command UUID generation/responde that we talked about before. In fact the nature of the PubSub system lends itself to those sorts of things when you have to do 'in-the-moment' verification of data (where something more akin to a webhook is appropriate). Technically we can skip the pubsub portion here and implement this all in the Service middleware, I think, so that's an option. Probably other strategies as well.

To be clear though this isn't certificate verification *per se* (we've been doing that for a while now by calling a SCEP-ish method (HasCN; which should be implemented in micromdm/scep but isn't yet). So I agree with you: validation whether the certificate (on its own) is valid should be done closer to the CA. But this PR is _specifically_ to address #358 where we need to assure the UDID that a client reports itself as having actually matches with the device certificate it used in its Authenticate message.

groob:
the pubsub system should be used to notify the various consumers of state changes, but for some uses RPC is much more preferable. IMO this applies here for auth. 

Thanks for the reminder to read the linked issue. That makes the scope of this PR much clearer.

platform/device/internal/deviceproto/device.proto
jessepeterson: We need to have an association from device record (UDID) to certificate, however. We probably don't need to store the whole cert. I'll likely refactor to just storing a hash of the certificate per the PR note.

mdm/server.go
jessepeterson: That's fine. As long as we shut down the rest of processing so a client can't access the command queue (through an HTTP error) that should do it.

mdm/mw/udidauth.go
jessepeterson: I'd be down with that. But if we're going that route then I'd suggest we change the very generic "builtin" package name we have duplicated on every service to be more in line with what it does. `builtinstorage`, `builtinpersistence`, `buildindb` or somesuch.

mdm/checkin_event.go
jessepeterson: It can be. The idea was that perhaps we'd want to access it somehow for other processing/usage? But that's fine — no need to overcomplicate for a use case we don't yet have.

jessepeterson:
Closing for #429.

PR #429, opened 2018-06-08
UDID/device certificate authentication (v.2)

### LINKED ISSUE #358: Match UDID to authentication

jessepeterson:
The rewrite of #428. No longer saves anything to the device store. Only stores `[]byte` of the certificate on the event (but does not persist). See #428 for further notes.

Resolves #358.


groob:
If you're validating against the cert store synchronously, is there a reason to pass the device cert to the event hook?

jessepeterson:
Are you suggesting keeping it in the Event messages/struct, but not marshaling it? I put it in the Event struct because that's the only abstraction available to service middleware that matches the service interface, no? However it seems like that'd be useful information to a webhook/pubsub user — having access to the device identity certificate for an API user's uses (whatever that might be).

groob:
Why would the dev certificate be useful to anyone beside the auth middleware?

jessepeterson:
The first thing that pops to mind is being able to encrypt profiles in a per-device manner. But I can imagine other uses. E.g. extended, custom validation of the certificate.

If it were to come out, though: what change do you propose? The only input you get that matches the service interfaces are the respective Event structs.

groob:
If you want to encrypt profiles with the device cert, you should ask the CA for the cert. Why send it with every request? 

I don't understand the second point. Struct fields are not part of the interface. Can you elaborate? 
All I'm proposing is removing the cert field from the event struct.

jessepeterson:
> If you want to encrypt profiles with the device cert, you should ask the CA for the cert. Why send it with every request?

If you're an API user you have no idea what certificate the device has. The CA won't know either, because the mapping of certificate <-> device is not a job of the CA. Depending on certificate attributes is not guaranteed nor even reliable (IMO).

> I don't understand the second point. Struct fields are not part of the interface.

Sure they are:

https://github.com/micromdm/micromdm/blob/41bb6456883f9e6d9ea990eede093bab5f07f708/mdm/service.go#L9-L12

`CheckinEvent` and `AcknowledgeEvent` are part of the service interface.

> Can you elaborate? All I'm proposing is removing the cert field from the event struct.

The event struct is the only thing passed into the Service, though (other than the context).

jessepeterson:
And, to be clear:

> Why send it with every request?

I do see the concern here. It's an extra 1K bytes (or however big a typical device cert is) along with every event. I'm not keen on that either. Perhaps the compromise is is to not sent it around with every Connect/Acknolwedge event, but keep it for Checkin events (which should be far more infrequent).

groob:
The only place we should care about the device cert is in the VerifySignature method. 
If we think we need the certificate for other reasons that need to be pushed down to the consumers, we can expose that in a future PR.

jessepeterson:
> The only place we should care about the device cert is in the VerifySignature method.

Er.. I'm not so sure.

The only thing VerifySignature does is 1) extract the device certificate from a base64-encoded CMS blob (i.e. the `Mdm-Signature` header) and 2) verify that that certificate is not bad/revoked (we do this by making sure the merely exists in the SCEP database).

The UDID/cert Middleware (this PR) is purely to address #358, which is about making sure the connection is not lying about its UDID. The only way to know that it isn't lying is to keep some additional state about which certificate it originally authenticated with (in its initial Checkin message). While we could add that additional body reading/Plist parsing to SignatureVerifier it's already done in the various `decodeRequests` (which output Event structs that then get handed to the Middleware).

> If we think we need the certificate for other reasons that need to be pushed down to the consumers, we can expose that in a future PR.

In this case this mdm.Service middleware here _**is**_ a "consumer" — in that it accepts the Event structs.

In fact thinking about it you refactor things such that you could ditch SignatureVerifier (instead considering certificate extract a "decoding" step) and actually implement the revocation checking as another chained mdm.Service middleware. I like the cleanliness and consistency of that model, in fact. But it would require keeping the cert in the Events.

But in this case: I suppose I can remove the fields from the protobufs (but keep them in the Event structs) if that helps. This would prevent any listeners from seeing that additional field (but still be able to use the Event structs themselves).

azherdev1978:
To add 2 cents, I was able to use device certificate to track the enrollment process from the initial download through the final token update by utilizing the Subject OU. Very clean in my case. I've ported the changes in the PR and it has really helped me out.

In fact, negated the need to use query parameters in the URL - in our case.

groob:
@jessepeterson I'm interested in the device cert events, but do you think we can separate the synchronous cert verification and the events into two PRs? Adding the certs to the events seems like a nice to have but not a requirement for resolving  #358.

jessepeterson:
@groob The more I think about it implementing the cert. revocation checking _also_ as an mdm Service middleware seems like a cleaner design.

That said we don't strictly need to provide the identity cert in the Events. I suppose we can modify the mdm Service interface to add that data being passed. Embedding it in the Events just seemed like a clean way to provide that data to other middleware layers _and_ give the device cert. to any pubsub subscriber as well.

To be clear what's your primary concerns against to putting it in the Events? Is it the amount of data/overhead? If so would it be okay putting either a cert. signature/hash in its place? Or maybe the requisite bits required for an OCSP request (serial #, issuer, etc.)

groob:
I don't have many concerns with adding it to the event, but I have a concern with relying on an async event to do a synchronous auth operation. 

So my concern here is that adding it to an even should be its own PR, and not part of the auth code.

jessepeterson:
> I don't have many concerns with adding it to the event, but I have a concern with relying on an async event to do a synchronous auth operation.

Sure, I share that same concern. The _first_ iteration of this code used async events (abandoned PR). But now that we're a full blown middleware that issues our own failure errors, we're back to fully synchronous. We're just using the Event to _store_ the certificate for our validation purposes in this case, before the publishing to pubsub is supposed to even happen.

> So my concern here is that adding it to an even should be its own PR, and not part of the auth code.

Roger that, that makes sense. It's just that we'll have to modify the mdm.Service interface to take it out of the Events, then un-modify it in the other PR to put it back into the Events. :)

jessepeterson:
Finally had some time to refactor this. I pulled out the certificate from the event and marshaled protobufs and instead modified the mdm service interface. I also got rid of the `SignatureVerifier` interface: since we now place the certificate in the middleware interface it's just another middleware that uses that data.

I particularly like this because it means that if you want to verify the certificate another way (say, OCSP against your own CA) you just need to implement a mw and handle the passed `Device` struct which contains the identity certificate. Also got rid of our duplicated `HasCN` (which was merged in to micromdm/scep a while back; we now get to use it because we're using a modern version of it).

platform/device/udidauth.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+
+	"github.com/micromdm/micromdm/mdm"
+)
+
+type UDIDCertAuthStore interface {
+	SaveUDIDCertHash(udid, cert_hash []byte) error
```
groob: use camel case.

platform/device/udidauth.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type udidCertAuthMiddleware struct {
+	store  UDIDCertAuthStore
+	next   mdm.Service
+	logger log.Logger
+}
+
+func hashCertRaw(c []byte) []byte {
+	ret_bytes := make([]byte, 32)
```
groob: snake case.

platform/device/udidauth.go
```diff
@@ -0,0 +1,98 @@
...
+	sum := sha256.Sum256(c)
+	copy(ret_bytes, sum[:])
+	return ret_bytes
+}
+
+func (mw *udidCertAuthMiddleware) validateUDIDCertAuth(udid, cert_hash []byte) (bool, error) {
+	db_cert_hash, err := mw.store.GetUDIDCertHash(udid)
```
groob: snake case.

mdm/acknowledge.go
```diff
@@ -6,12 +6,11 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/endpoint"
-	"github.com/groob/plist"
 	"github.com/pkg/errors"
 	uuid "github.com/satori/go.uuid"
 )
 
-func (svc *MDMService) Acknowledge(ctx context.Context, req AcknowledgeEvent) (payload []byte, err error) {
+func (svc *MDMService) Acknowledge(ctx context.Context, req AcknowledgeEvent, _ Device) (payload []byte, err error) {
```
groob: I would prefer to see the Certificate either as fields on `AcknowledgeEvent`/`CheckinEvent` or as a context value. I don't see a reason we should add the new param. It's only used in middleware, and would be ignored by the actual implementation, so it would be cleaner not to force it in the function signature.

jessepeterson: I chose to put it on the context. I don't feel good about putting elements on the events without also having them marshalled. Just seems like a hidden/undiscoverable thing.

PR #440, opened 2018-06-14
Use micromdm/scep master (for newer pkcs7 lib)

### LINKED ISSUE #436: Error enrolling on macOS 10.12.6

v1.3.0
commit 25a07b82b5407a1115b6cf6770716cfdc43c9719

Attempted to enroll from a macOS 10.12.6 machine. The server was running through ngrok using the ngrok instructions. The default enrollment profile was being used and a blueprint was set to push out two profiles

The machine to enroll in mdm

### What did you see instead?

A Profile installation failed error from macOS on the client machine. The profile could not be installed due to an unexpected error. <NSOSStatusErrorDomain: -603

In the console logs of the client there is this additional info

```
[ERROR] <: [MDM_SCEP_Enroll] Calling SCEPRequestCertSignature --> <NSOSStatusErrorDomain:-603>
```

and the following on the server

```
transport=http method=GET status=200 proto=HTTP/1.1 host=::1 user_agent="mdmclient/960 CFNetwork/811.9 Darwin/16.7.0 (x86_64)" path="/scep?operation=GetCACert&message=Device%20Management%20Identity%20Certificate"
method=PKIOperation err="invalid data len 0" took=4.216921ms
invalid data len 0
transport=http err="invalid data len 0"
transport=http method=POST status=500 proto=HTTP/1.1 host=::1 user_agent="mdmclient/960 CFNetwork/811.9 Darwin/16.7.0 (x86_64)" path="/scep?operation=PKIOperation"
```

So it looks like the POST to /scep is returning a 500 Internal Server Error for the reason of `invalid data len 0`

jessepeterson:
micromdm/scep#87 is still open for that project. But this PR should resolve #436 (as we're compiling against the fixed version of pkcs7 in @groob's branch). I've tested with a 10.12.6 VM and it seems to enroll fine, but could definitely use some testing.

I'm not super sure about the Handler changes. I feel there's a better way to do this, especially with the doubled-up path Handlers in both the scepserver MakeHandlers, and the r.Handler in micromdm. Also - the `dep` tool stuff is a bit of a nightmare, if folks can test to make sure they're `dep ensure` works fine that'd be great, too.

Feedback welcome!

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -18,10 +18,14 @@
   version = "v1.3.1"
 
 [[projects]]
-  branch = "master"
+  branch = "sha256"
   name = "github.com/fullsailor/pkcs7"
-  packages = ["."]
-  revision = "ae226422660e5ca10db350d33f81c6608f3fbcdd"
+  packages = [
+    ".",
+    "internal/x509util"
+  ]
+  revision = "f41fa115eb76823dcfb8c45678da0fb6d7ecb792"
+  source = "github.com/groob/pkcs7"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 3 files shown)


knightsc:
I tested on 10.12.6 and it seems to be working there as well

jessepeterson:
Thanks for testing, folks! That's most appreciated! I'll clean-up a couple comments and stuff then merge.

jessepeterson:
@groob what's your opinion on using the `EndpointLoggingMiddleware` in micromdm?

PR #442, opened 2018-06-16
Add software update command examples

knightsc:
According to the [Apple documentation](https://developer.apple.com/library/archive/documentation/Miscellaneous/Reference/MobileDeviceManagementProtocolRef/3-MDM_Protocol/MDM_Protocol.html#//apple_ref/doc/uid/TP40017387-CH3-SW302) "On macOS, all supported Software Update commands except the AvailableOSUpdates query require DEP enrollment."

So far I've tested these changes on macOS 10.13.4. And all commands except the `ScheduleOSUpdate` command seemed to work even though my test machine isn't DEP enrolled. Attempting to run a `ScheduleOSUpdate` command will fail for reason of not DEP enrolled on a non DEP machine.

Here's some examples of using the example commands:

* `./tools/api/commands/schedule_os_update_scan <UDID> <force (true|false)`
* `./tools/api/commands/available_os_updates <UDID>`
* `./tools/api/commands/schedule_os_update <UDID> <ProductKey> <InstallAction>`
* `./tools/api/commands/os_update_status <UDID>`

I have not tried these out on iOS yet.


mdm/mdm/command.go
```diff
@@ -56,6 +56,7 @@ type Command struct {
 	SetAutoAdminPassword            *SetAutoAdminPassword
 	ScheduleOSUpdate                *ScheduleOSUpdate
 	ScheduleOSUpdateScan            *ScheduleOSUpdateScan
+	OSUpdateStatus                  *OSUpdateStatus
```
groob: This is not necessary to add a struct since it's `RequestType` with no additional keys.

mdm/mdm/internal/mdmproto/mdm.proto
```diff
@@ -64,6 +65,7 @@ message Command {
     ScheduleOSUpdateScan schedule_os_update_scan = 33;
     ActiveNSExtensions active_ns_extensions = 34;
     RotateFileVaultKey rotate_filevault_key = 35;
+    OSUpdateStatus os_update_status = 36;
```
groob: not necessary to create the status message

groob:
Remove the Go Struct/Proto Message and this is good to merge. Thanks a lot for adding the PR!

groob:
Thanks! Awesome work.

PR #444, opened 2018-06-20
Allow removal of device records

### LINKED ISSUE #394: Non-functioning Device Removal
If a device is permanently damaged and can not be removed via un-enrollment via `mdmctl apply block` method, there is no way to remove that device from the MDM. A command to nuke device from the database.

knightsc:
This adds a `remove devices` command to `mdmctl`. You can pass a single UDID or a comma separated list of UDIDs. Only the device records are removed.

Resolves #394


groob:
This PR nice and clean, I like all the changes. But I think it could be two separate PRs. If you don't mind can you put up the diff for creating the `apns` worker as a separate pull request?

knightsc:
Yup, will do

knightsc:
Moved the push worker commit over to #445

PR #446, opened 2018-06-21
Create a custom http.Client for apns communication

### LINKED ISSUE #215: push api loses active connection after a while
If I use the `/push` endpoint to trigger a push notification 

```
{
  "error": "Post https://api.push.apple.com/3/device/[redacted]: read tcp [redacted]: read: connection reset by peer"
}
```

sending it a second time causes messages to succeed after. 

This might be an internal thing we're not being careless about, or something to look at in bufford. 

needs investigation

knightsc:
In order to set custom transport options on the http.Client used
by the push service we need to create our own instance. This
allows us to configure the IdleConnTimeout and prevent connections
from being killed and pushes not being sent. 90 seconds is used
here since it's the same as the DefaultTransport provided by Go.

Fixes #215

### DIFF
--- a/platform/apns/service.go
+++ b/platform/apns/service.go
@@ -5,10 +5,13 @@ import (
 	"crypto/tls"
 	"fmt"
 	"log"
+	"net/http"
 	"sync"
+	"time"
 
 	"github.com/RobotsAndPencils/buford/push"
 	"github.com/pkg/errors"
+	"golang.org/x/net/http2"
 
 	"github.com/micromdm/micromdm/platform/config"
 	"github.com/micromdm/micromdm/platform/pubsub"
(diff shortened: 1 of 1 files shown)


knightsc:
@azherdev1978 can you re-run your test from #215 on this branch and see if it fixes your issue?

platform/apns/service.go
```diff
@@ -121,13 +124,30 @@ func updateClient(svc *PushService, sub pubsub.Subscriber) error {
...
+	transport := &http.Transport{
+		TLSClientConfig: config,
+		IdleConnTimeout: 90 * time.Second,
+	}
+
+	if err := http2.ConfigureTransport(transport); err != nil {
+		return nil, err
+	}
+
+	return &http.Client{Transport: transport}, nil
```
groob: the `http.Client` field also has a timeout which covers a few other places in the request that could time. out. Lets set this to 
```
return &http.Client{
  Transport: transport,
  Timeout: 20 * time.Second,
}, nil
```

knightsc: Added in 021962d

PR #447, opened 2018-06-22
Cleanup blueprint indexing

knightsc:
Saving a blueprint was creating multiple records in the blueprint index bucket. One was the index that gets used for lookup which is a key of the blueprint Name and the value\ being the blueprint UUID and the other record had a key and value of blueprint UUID. The second with a key of UUID is not needed since the key in the main blueprint bucket is already UUID

My first commit creates some basic unit tests of the blueprint DB. It does not currently exercise all the functionality like adding profiles, applications, users to the blueprint. But I tried to make sure the save and lookup functions were being tested since that's the code I'm changing here.

Additionally it looked like the Create method in the blueprint DB was unused and not part of the Store definition

```
type Store interface {
	Save(*Blueprint) error
	BlueprintByName(name string) (*Blueprint, error)
	List() ([]Blueprint, error)
	Delete(string) error
}
```

These are the first real set of Go unit tests I've written so if stylistically there's a more Go appropriate way to write some of these let me know.

### DIFF
--- a/platform/blueprint/builtin/db.go
+++ b/platform/blueprint/builtin/db.go
@@ -47,17 +47,6 @@ func NewDB(
 	return datastore, nil
 }
 
-func (db *DB) Create(bp *blueprint.Blueprint) error {
-	_, err := db.BlueprintByName(bp.Name)
-	if err != nil && isNotFound(err) {
-		return errors.New("blueprint must have a unique name")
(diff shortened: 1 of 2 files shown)


groob:
I think this is good but going to flag @jessepeterson for a review/sanity check.

PR #455, opened 2018-07-12
Allow setting curl options in environment variable

jessepeterson:
Also fix get devices requiring some body content.


knightsc:
@jessepeterson Just out of curiosity, what are some things I might want to set via `$CURL_OPTS`?

jessepeterson:
@knightsc the main one is `-k` for self-signed certs. But I imagine you'd want to debug the headers with `-v` and such.

PR #480, opened 2018-08-06
Use config for block push

### LINKED ISSUE #479: TLS error when using mdmctl apply block with self signed certificates
1.3.1

mdmctl apply block -udid=UUID

device checks in, gets 401, unenrolls

### What did you see instead?
```
method=BlockDevice udid=UUID err=null took=422.687µs
transport=http method=POST status=200 proto=HTTP/1.1 host=::1 user_agent=Go-http-client/1.1 path=/v1/devices/UUID/block
ts="2018/08/06 16:19:27" caller="http: TLS handshake error from [::1" msg="]:57551: remote error: tls: bad certificate"
```

After talking with @jessepeterson the interim solution was to send a push notification to force a checkin. I have put some notes [on the wiki](https://github.com/micromdm/micromdm/wiki/Using-MicroMDM-with-self-signed-certificates#self-certs-and-non-removable-dep-profiles)

From jesse:

```
so we're sending a push immediately after the block request, but we're not using our http config for that purpose - thus bypassing the self-signed trust settings for mdmctl

so that TLS error is just when its trying to send a push after the block has been set

it's an artifact that we don't have `mdmctl` support for sending a push - that's an API-only endpoint 

we should (at the least) use our already configurd http settings
```

jessepeterson:
Fixes #479

### DIFF
--- a/cmd/mdmctl/apply.go
+++ b/cmd/mdmctl/apply.go
@@ -225,7 +225,7 @@ func (cmd *applyCommand) applyBlock(args []string) error {
 		return nil
 	}
 	req.SetBasicAuth("micromdm", cmd.config.APIToken)
-	http.DefaultClient.Do(req)
+	skipVerifyHTTPClient(cmd.config.SkipVerify).Do(req)
 	return nil
 }
 


jessepeterson:
Thanks for the review! :)

PR #489, opened 2018-08-14
Support an empty body

jessepeterson:
Support an empty body (i.e. don't require JSON) if the content-length is 0 for the device list.

### DIFF
--- a/platform/device/get_devices.go
+++ b/platform/device/get_devices.go
@@ -49,7 +49,11 @@ func (r getDevicesResponse) Failed() error { return r.Err }
 
 func decodeListDevicesRequest(ctx context.Context, r *http.Request) (interface{}, error) {
 	var opts ListDevicesOption
-	err := httputil.DecodeJSONRequest(r, &opts)
+	var err error
+	// support an empty GET request without a body
+	if r.ContentLength > 0 {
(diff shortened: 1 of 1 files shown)


jessepeterson:
This has a regression: if the client doesn't send a content-length then I think this will always skip the json decoding.

knightsc:
@jessepeterson So what do you think the solution is? I'm sure we could make mdmctl set the Content-Length header. But I would imagine most people trying out this endpoint via postman or curl won't and in turn they're more likely to get a list of devices with an empty body but have things break later when they try to use the filter.

Is it GoKit's convention to simply encode all json as the body of all requests regardless of HTTP method?

groob:
IMO we should just change this method to a POST, or consider changing filtering to url params instead of a JSON body.

jessepeterson:
To _fix_ this we'd need to actually implement a body reader that tries to get content unconditionally (either buffering it, or to be efficient just read the first bit to try and assess EOF and hand off to the JSON reader).

But that's a lot of sillyness just to make an optional empty GET request. Per past discussion in Slack I agree with @groob — it should just become a POST or move to URL params (my vote is the former). Body-in-GET isn't a great convention as it is. Was trying to just 'quick'-fix this but I forgot about the streaming/indeterminate nature of HTTP requests.

PR #490, opened 2018-08-18
Remove SCEP CA certificate from enrollment profile & disk

jessepeterson:
The SCEP CA certificate is not needed to establish trust between the client and MDM server. Thus including it in the enrollment profile is unnecessary. Separately, and since we don't need it, we can clean up the enrollment service a little by not reading the certificate from disk. Finally — just don't write it out to disk at all. Its usefulness on disk (for the sake of being in its own file) is debatable. Does anybody use it? Besides, anybody needing it can just query `/scep?operation=GetCACert` to retrieve it.

This should probably get wide testing across various device types — MicroMDM has always included this certificate in the enrollment profile. This change worked to enroll a macOS 10.12.6 host.

### DIFF
--- a/mdm/enroll/service.go
+++ b/mdm/enroll/service.go
@@ -29,17 +29,9 @@ type Service interface {
 	OTAPhase3(ctx context.Context) (profile.Mobileconfig, error)
 }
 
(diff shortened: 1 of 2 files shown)


jessepeterson:
Wow, @knightsc! Thanks for that very exhaustive testing!

PR #492, opened 2018-08-23
Correct SCEP profile payload

knightsc:
Both KeyType and KeyUsage were being defined with spaces in the name.
According to the spec this is incorrect. The clients were most likely just
ignoring these fields and everything still worked but it makes sense to correct
the profile.


jessepeterson:
The new docs seem to have spaces again: https://developer.apple.com/documentation/devicemanagement/scep/payloadcontent#properties

PR #497, opened 2018-08-28
Re-fetch on invalid cursor

jessepeterson:
Re-try fetch after invalid cursor

### DIFF
--- a/dep/depsync/depsync.go
+++ b/dep/depsync/depsync.go
@@ -177,6 +177,10 @@ func isCursorExpired(err error) bool {
 	return strings.Contains(err.Error(), "EXPIRED_CURSOR")
 }
 
+func isCursorInvalid(err error) bool {
+	return strings.Contains(err.Error(), "INVALID_CURSOR")
+}
+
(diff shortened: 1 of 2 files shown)


dep/depsync/depsync.go
```diff
@@ -287,6 +292,15 @@ FETCH:
 		resp, err := w.client.FetchDevices(dep.Limit(100), dep.Cursor(w.conf.Cursor.Value))
 		if err != nil && isCursorExhausted(err) {
 			goto SYNC
+		} else if err != nil && isCursorInvalid(err) {
```
groob: maybe time to convert this error check to a `switch`?

jessepeterson: yeah that'd be a good idea. i was thinking of changing the sync loop to be a single loop for both fetch and sync (rather than the separate `goto` calls). a lot of the logging and handling code is repeated. i'll probably submit a depsync refactor PR at some point - i also want to change the autoassigner to be a 'worker' powered off the pubsub rather than directly called goroutines.

PR #503, opened 2018-09-07
Retrieve certificate

jessepeterson:
Allow reading the MDM push cert from the server. To e.g. periodically check its expiration or to idempotently update the push cert.


groob:
I'm ok with this.

PR #517, opened 2018-10-02
Some cleanup based on `goreportcard` linters

daniellockard:


cmd/mdmctl/config.go
```diff
@@ -118,7 +118,10 @@ func migrateServerConfig(configName string) error {
 	}
 	err = switchServerConfig(configName)
 	if err != nil {
-		fmt.Errorf("Failed to set %s as active config", configName)
```
groob: nice catch!

cmd/mdmctl/config.go
```diff
@@ -118,7 +118,10 @@ func migrateServerConfig(configName string) error {
...
+		err = fmt.Errorf("Failed to set %s as active config", configName)
+		if err != nil {
+			return err
```
groob: `return errors.Wrapf(err, "Failed to set %s as active config", configName)`

platform/device/udidauth.go
```diff
@@ -102,5 +102,4 @@ func (mw *udidCertAuthMiddleware) Checkin(ctx context.Context, req mdm.CheckinEv
 	default:
 		return errors.Errorf("unknown checkin message type %s", req.Command.MessageType)
 	}
-	return mw.next.Checkin(ctx, req)
```
jessepeterson: How else is the next middleware layer going to be called?

groob: it's returned in the case statements. this is dead code anyway because of the `default`. 

since each switch case has a return, the last line was dead code.

PR #518, opened 2018-10-02
Error Checks and Dead Code Removal

daniellockard:
I'm sorry this is so huge - I know it might take a bit to review. I tried to follow the code style of what was around these calls to either return the error, or print it.  Part of the issue is that the `Log` calls all can return an Error, and there are cases where we're using the logger to log an error, in that case I think it's best to Println.

### DIFF
--- a/cmd/mdmctl/apply.go
+++ b/cmd/mdmctl/apply.go
@@ -42,7 +42,10 @@ func (cmd *applyCommand) setup() error {
 
 func (cmd *applyCommand) Run(args []string) error {
 	if len(args) < 1 {
-		cmd.Usage()
+		err := cmd.Usage()
+		if err != nil {
+			return err
+		}
 		os.Exit(1)
 	}
 	if err := cmd.setup(); err != nil {
@@ -67,7 +70,10 @@ func (cmd *applyCommand) Run(args []string) error {
 	case "dep-autoassigner":
 		run = cmd.applyDEPAutoAssigner
 	default:
-		cmd.Usage()
+		err := cmd.Usage()
+		if err != nil {
+			return err
+		}
 		os.Exit(1)
 	}
 	return run(args[1:])
@@ -215,18 +221,16 @@ func (cmd *applyCommand) applyBlock(args []string) error {
 	// trigger a push
 	u, err := url.Parse(cmd.config.ServerURL)
 	if err != nil {
-		fmt.Println(err)
-		return nil
+		return err
 	}
 	u.Path = "/push/" + url.QueryEscape(*flUDID)
 	req, err := http.NewRequest("GET", u.String(), nil)
 	if err != nil {
-		fmt.Println(err)
-		return nil
+		return err
 	}
 	req.SetBasicAuth("micromdm", cmd.config.APIToken)
-	skipVerifyHTTPClient(cmd.config.SkipVerify).Do(req)
-	return nil
+	_, err = skipVerifyHTTPClient(cmd.config.SkipVerify).Do(req)
+	return err
 }
 
 func (cmd *applyCommand) applyProfile(args []string) error {
--- a/cmd/mdmctl/mdmcert.download.go
+++ b/cmd/mdmctl/mdmcert.download.go
@@ -96,13 +96,19 @@ func (cmd *mdmcertDownloadCommand) Run(args []string) error {
 	)
 
 	if err := flagset.Parse(args); err != nil {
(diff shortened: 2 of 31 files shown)


cmd/mdmctl/mdmcert.download.go
```diff
@@ -213,7 +219,10 @@ func (cmd *mdmcertDownloadCommand) Run(args []string) error {
 		if err != nil {
 			return errors.Wrap(err, "reading PKI private key")
 		}
-		ioutil.WriteFile("/tmp/fubar.p7", pkcsBytes, 0666)
+		err = ioutil.WriteFile("/tmp/fubar.p7", pkcsBytes, 0666)
```
jessepeterson: This looks like never-removed debugging code. Not sure why we'd need to write this out.

daniellockard: I doubt we do :), especially with a name like "fubar"

cmd/mdmctl/mdmcert.go
```diff
@@ -178,7 +184,10 @@ func (cmd *mdmcertCommand) runPush(args []string) error {
 	}
 
 	if err := os.MkdirAll(filepath.Dir(*flCSRPath), 0755); err != nil {
-		errors.Wrapf(err, "create directory %s", filepath.Dir(*flCSRPath))
+		innerErr := errors.Wrapf(err, "create directory %s", filepath.Dir(*flCSRPath))
+		if innerErr != nil {
+			return innerErr
```
jessepeterson: any reason for not just returning the Wrapf here?

cmd/mdmctl/apply.go
```diff
@@ -42,7 +42,10 @@ func (cmd *applyCommand) setup() error {
 
 func (cmd *applyCommand) Run(args []string) error {
 	if len(args) < 1 {
-		cmd.Usage()
+		err := cmd.Usage()
```
groob: would prefer to see single returns on the same line as the if
` if err := (); err != nil ...`

cmd/mdmctl/mdmcert.download.go
```diff
@@ -96,13 +96,19 @@ func (cmd *mdmcertDownloadCommand) Run(args []string) error {
...
+		if innerErr != nil {
+			return errors.Wrap(innerErr, "calling usage")
+		}
+		return errors.Wrap(err, "failed parsing")
 	}
 
 	// neither flag was used
 	if !*flNew && *flDecrypt == "" {
-		cmd.Usage()
+		innerErr := cmd.Usage()
```
groob: if err := cmd.Usage(); err != nil {

mdm/appmanifest/appmanifest_test.go
```diff
@@ -39,7 +39,10 @@ func benchmarkSize(b *testing.B, size int) {
 	r := bytes.NewReader(buf)
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		calculateMD5s(r, DefaultMD5Size)
+		_, err := calculateMD5s(r, DefaultMD5Size)
+		if err != nil {
+			return
```
groob: b.Fatal(err)

pkg/httputil/httputil.go
```diff
@@ -69,7 +69,10 @@ func ErrorEncoder(_ context.Context, err error, w http.ResponseWriter) {
 	}
 	w.WriteHeader(code)
 
-	enc.Encode(errMap)
+	err = enc.Encode(errMap)
+	if err != nil {
+		fmt.Println(errors.Wrap(err, "encoding errMap"))
```
groob: need to either use the logger or write a 500. I think not checking the error here is preferable. Failing to encode just means we can't write back to the client for some reason.

platform/apns/worker.go
```diff
@@ -45,10 +46,13 @@ func (w *Worker) Run(ctx context.Context) error {
 			err = w.updatePushInfoFromTokenUpdate(ctx, event.Message)
 		}
 		if err != nil {
-			level.Info(w.logger).Log(
+			innerErr := level.Info(w.logger).Log(
```
groob: logger should not be checked for error. the error is only used for audit logging.
FYI, `fmt.Print` returns an error too. 
https://golang.org/pkg/fmt/#Println

daniellockard: Ah, you're not wrong. I'll rip out all the logger error checking.

PR #521, opened 2018-10-09
add support for the ApplicationConfiguration setting

groob:
wip branch, This adds the setting type and the `identifier` field, but the Configuration field is a bit TBD. I'll get my hand on an iOS device and complete it


groob:
The marshaling to the plist isnt quite working yet. I'll poke at it in the evening again to see why.

PR #526, opened 2018-10-10
GET to POST for v1/users endpoint

jenjac:
Changed GET to POST as a GET with a request body is not preferred.


groob:
Is there a reason these all have to be separate PRs?

PR #538, opened 2018-10-18
Adding a VPP Package

jenjac:
Currently includes methods to create a vpp client with an sToken, set client context, and associate and disassociate serials to and from vpp apps.

### DIFF
--- a/vpp/clientconfigsrv.go
+++ b/vpp/clientconfigsrv.go
@@ -1,0 +1,117 @@
+package vpp
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/pkg/errors"
+	"github.com/satori/go.uuid"
+	"strings"
+)
+
+type ClientContext struct {
+	HostName string `json:"hostname"`
+	Guid     string `json:"guid"`
+}
+
+type Location struct {
+	LocationName string `json:"locationName"`
+	LocationId   int    `json:"locationId"`
+}
+
+type ClientConfigSrvResponse struct {
+	ClientContext      interface{} `json:"clientContext"`
+	AppleId            string      `json:"appleId"`
+	OrganizationIdHash string      `json:"organizationIdHash"`
+	Status             int         `json:"status"`
+	OrganizationId     int         `json:"organizationId"`
+	UId                string      `json:"uId"`
+	CountryCode        string      `json:"countryCode"`
+	Location           Location    `json:"location"`
+	ApnToken           string      `json:"apnToken"`
+	Email              string      `json:"email"`
+}
+
+type ClientConfigSrv struct {
+	ClientContext      ClientContext `json:"clientContext"`
+	AppleId            string        `json:"appleId"`
+	OrganizationIdHash string        `json:"organizationIdHash"`
+	Status             int           `json:"status"`
+	OrganizationId     int           `json:"organizationId"`
+	UId                string        `json:"uId"`
+	CountryCode        string        `json:"countryCode"`
+	Location           Location      `json:"location"`
+	ApnToken           string        `json:"apnToken"`
+	Email              string        `json:"email"`
+}
+
+func (c *Client) GetClientConfigSrv() (*ClientConfigSrv, error) {
+	request := map[string]interface{}{
+		"sToken": c.sToken,
+	}
+
+	clientConfigSrvUrl := c.VPPServiceConfigSrv.ClientConfigSrvUrl
+	var response ClientConfigSrvResponse
+	req, err := c.newRequest("POST", clientConfigSrvUrl, request)
+	if err != nil {
+		return nil, errors.Wrap(err, "create ClientConfigSrv request")
+	}
+
+	err = c.do(req, &response)
+
+	var context = response.ClientContext
+	var clientContext ClientContext
+	json.NewDecoder(strings.NewReader(context.(string))).Decode(&clientContext)
+
+	config := ClientConfigSrv{
+		ClientContext:      clientContext,
+		AppleId:            response.AppleId,
+		OrganizationIdHash: response.OrganizationIdHash,
+		Status:             response.Status,
+		OrganizationId:     response.OrganizationId,
+		UId:                response.UId,
+		CountryCode:        response.CountryCode,
+		Location:           response.Location,
+		ApnToken:           response.ApnToken,
+		Email:              response.Email,
+	}
+	return &config, errors.Wrap(err, "ClientConfigSrv request")
+}
+
+func (c *Client) GetClientContext() (*ClientContext, error) {
+	clientConfigSrv, err := c.GetClientConfigSrv()
+	if err != nil {
+		return nil, errors.Wrap(err, "create ClientConfigSrv request")
+	}
+
+	var context = clientConfigSrv.ClientContext
+	return &context, nil
+}
+
+func (c *Client) SetClientContext(serverURL string) (*ClientContext, error) {
+	uuid, err := uuid.NewV4()
+	if err != nil {
+		return nil, errors.Wrap(err, "create uuid")
+	}
+
+	newContext := map[string]interface{}{
+		"sToken":        c.sToken,
+		"clientContext": fmt.Sprintf("{\"hostname\":\"%s\",\"guid\":\"%s\"}", serverURL, uuid),
+	}
+
+	clientConfigSrvUrl := c.VPPServiceConfigSrv.ClientConfigSrvUrl
+
+	var response ClientConfigSrvResponse
+
+	req, err := c.newRequest("POST", clientConfigSrvUrl, newContext)
+	if err != nil {
+		return nil, errors.Wrap(err, "create Client Context ClientConfigSrv request")
+	}
+
+	err = c.do(req, &response)
+
+	var context = response.ClientContext
+	var clientContext ClientContext
+	json.NewDecoder(strings.NewReader(context.(string))).Decode(&clientContext)
+
+	return &clientContext, errors.Wrap(err, "set Client Context ClientConfigSrv request")
+}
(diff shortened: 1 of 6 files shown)


vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+type Location struct {
+	LocationName string `json:"locationName"`
+	LocationId   int    `json:"locationId"`
+}
+
+type ClientConfigSrvResponse struct {
+	ClientContext      interface{} `json:"clientContext"`
+	AppleId            string      `json:"appleId"`
```
groob: a good pattern to keep in mind is to keep initialisms all upper case. `AppleID`,`UID`, `APNToken`, `GUID` etc. 

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,117 @@
...
+	OrganizationId     int           `json:"organizationId"`
+	UId                string        `json:"uId"`
+	CountryCode        string        `json:"countryCode"`
+	Location           Location      `json:"location"`
+	ApnToken           string        `json:"apnToken"`
+	Email              string        `json:"email"`
+}
+
+func (c *Client) GetClientConfigSrv() (*ClientConfigSrv, error) {
+	request := map[string]interface{}{
```
groob: create request and response structs for each action. 
If you want to keep the request as an anonymous type/variable thats ok. See here as an example https://github.com/micromdm/micromdm/blob/a230831725f173f779cf7e030a93a71b32e8ab13/dep/devices.go#L108-L112

vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	clientConfigSrvUrl := c.VPPServiceConfigSrv.ClientConfigSrvUrl
+	var response ClientConfigSrvResponse
+	req, err := c.newRequest("POST", clientConfigSrvUrl, request)
+	if err != nil {
+		return nil, errors.Wrap(err, "create ClientConfigSrv request")
+	}
+
+	err = c.do(req, &response)
```
groob: if err := c.do...;err ! = nil { return } 
Always check errors

vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,117 @@
...
+
+	var context = response.ClientContext
+	var clientContext ClientContext
+	json.NewDecoder(strings.NewReader(context.(string))).Decode(&clientContext)
```
groob: check error

vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,117 @@
...
+	clientConfigSrvUrl := c.VPPServiceConfigSrv.ClientConfigSrvUrl
+	var response ClientConfigSrvResponse
+	req, err := c.newRequest("POST", clientConfigSrvUrl, request)
+	if err != nil {
+		return nil, errors.Wrap(err, "create ClientConfigSrv request")
+	}
+
+	err = c.do(req, &response)
+
+	var context = response.ClientContext
```
groob: can you explain clientContext to me? Its not quite clear why it's named this way and what it does.

I see below you happen to sprintf a context too. Not sure I follow.

vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,117 @@
...
+	Guid     string `json:"guid"`
+}
+
+type Location struct {
+	LocationName string `json:"locationName"`
+	LocationId   int    `json:"locationId"`
+}
+
+type ClientConfigSrvResponse struct {
+	ClientContext      interface{} `json:"clientContext"`
```
groob: why is this an empty interface?

vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (c *Client) SetClientContext(serverURL string) (*ClientContext, error) {
+	uuid, err := uuid.NewV4()
+	if err != nil {
+		return nil, errors.Wrap(err, "create uuid")
+	}
+
+	newContext := map[string]interface{}{
+		"sToken":        c.sToken,
+		"clientContext": fmt.Sprintf("{\"hostname\":\"%s\",\"guid\":\"%s\"}", serverURL, uuid),
```
groob: very confused by clientContext. Why sprintf and escape strings instead of creating a struct?

vpp/getvppassetssrv.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "create VppAssetsSrv request")
+	}
+	var assets = response.Assets
+
+	var pricing string
+	for i := 0; i < len(assets); i++ {
+		asset := assets[i]
+		if asset.AdamIdStr == appId {
+			pricing = asset.PricingParam
```
groob: do you need to `break` here?

vpp/clientconfigsrv.go
jenjac: I believe I have address these in the commit that I just pushed

vpp/clientconfigsrv.go
jenjac: I believe I have address these in the commit that I just pushed

vpp/clientconfigsrv.go
jenjac: I am not sure how to check an error from the json.NewDecoder function as it just assigns the decoded value to the clientContext variable

vpp/clientconfigsrv.go
jenjac: I removed the sprintf line.
ClientContext is a string returned in the json from the ClientConfigSrv.  It associates a particular VPP token with your particular MDM.
Practically, if you send an InstallApplication command from your MDM to a device for a VPP app, the app won't install unless the ClientContext set in the VPP token matches the URL of your MDM server.

vpp/clientconfigsrv.go
jenjac: I changed it to a string.  I then convert that string to a ClientContext struct only when needed.  In this case in the SetClientContext and GetClientContext methods.

vpp/clientconfigsrv.go
jenjac: I changed this to use the ClientContext struct to create a new ClientContext, converting the struct to the string that is needed.

vpp/getvppassetssrv.go
jenjac: I added the break

vpp/clientconfigsrv.go
jenjac: This err is returned to be checked by the calling function.  I modeled it after "dep/device.go" lines 119 and 120. Is there something else I should be doing?

vpp/getlicensessrv.go
```diff
@@ -0,0 +1,88 @@
...
+// Checks if a particular serial is associated with an appID
+func (c *Client) CheckAssignedLicense(serial string, appID string) (bool, error) {
+	// Get all licenses with serial associated
+	licenses, err := c.GetLicensesForSerial(serial)
+	if err != nil {
+		return false, errors.Wrap(err, "create LicensesSrv request")
+	}
+
+	// Check for the particular appID
+	for i := 0; i < len(licenses); i++ {
```
groob: you can also range over with `for _, lic := range licenses`

jenjac: done in latest commit

vpp/getlicensessrv.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Gets licenses with specified serial associated
+func (c *Client) GetLicensesForSerial(serial string) ([]License, error) {
+	options := GetLicensesSrvOptions{
+		SerialNumber: serial,
+	}
+
+	response, err := c.GetLicensesSrv(options)
+	licenses := response.Licenses
```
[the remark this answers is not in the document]
jenjac: fixed in latest commit

vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,147 @@
...
+	response, err := c.GetClientConfigSrv(options)
+	if err != nil {
+		return nil, errors.Wrap(err, "create ClientConfigSrv request")
+	}
+
+	// Get the new ClientContext string
+	var contextString = response.ClientContext
+	// Convert the string to a ClientContext type
+	var clientContext ClientContext
+	json.NewDecoder(strings.NewReader(contextString)).Decode(&clientContext)
```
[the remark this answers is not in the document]
jenjac: fixed in last commit

vpp/clientconfigsrv.go
jenjac: fixed in last commit

jenjac:
ClientContext is definitely a challenge. I am pretty sure that it is received from Apple within the ClientConfigSrv response json as a string that looks like a json.  I tried receiving it directly into a ClientContext struct within the ClientConfigSrv struct, but I only get errors.  And if I am trying to set the ClientContext, it only works if I set that value as a string.

I have seen this before in the responses that I get from the command-webhook.  The postData.contents value seems to be a string that looks like a json as well.  I had resolved issues working with that in my Google Apps Script interface first.  That is how I recognized the same issue in working with ClientContext and came up with the solution that I did.

It would be great to get the Library merged at this point and I will be happy to work on a PR for integrating it into the server.

Where can I look for an example about how I might go about doing that?

jenjac:
Also, should I be doing something to resolve the "All checks have failed" message?

groob:
if you click on the Details link, it takes you to the test failure:

```
vpp/clientconfigsrv.go:122:12: assignment mismatch: 2 variables but 1 values
```

If you run `go install` in the `vpp` directory you will get the same failure.

vpp/clientconfigsrv.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "get ClientContext request")
+	}
+
+	// Get the ClientContext string
+	var context = clientConfigSrv.ClientContext
+
+	// Convert the string to a ClientContext type
+	var clientContext ClientContext
+	decodeError := json.NewDecoder(strings.NewReader(context)).Decode(&clientContext)
```
groob: just keep `err = `, it's very rare to ever need a new error variable.

jenjac: I posted a commit for this, just this morning.

groob:
This is great. Thanks for bearing with me on the code review!

PR #539, opened 2018-10-21
Profile Management workflow 

groob:
TODO: delete device profiles which got removed on the device.


GavinElder:
If the delete behaviour is the only thing left to implement is it worth landing this and picking it up in a future PR.

Issue #270, opened 2017-10-27
mdmctl should be more forgiving about singular/plural


groob:
```
mdmctl get device
mdmctl get devices
mdmctl get dev
```
These three commands should do the same thing even if `devices` is the only officially documented command.

PR #540, opened 2018-10-25
Allow singulars for devices

emman27:
Resolves https://github.com/micromdm/micromdm/issues/270


cmd/mdmctl/get.go
```diff
@@ -56,6 +54,10 @@ func (cmd *getCommand) Run(args []string) error {
 	switch strings.ToLower(args[0]) {
 	case "devices":
 		run = cmd.getDevices
+	case "dev":
```
groob: Case can be comma separated:
```
case “foo”,”bar”,”baz”:
```

Use that instead of creating new cases.

emman27:
Whoa I never knew that. Thanks!

PR #561, opened 2019-01-23
Add Docs from the Wiki with slight change.

gavinelder:
# Background

Verbatim copy of the docs. 

Removal of sidebar, change welcome to readme.md 

Begin folder structure.


groob:
I'm 👍 on this change, but sounds like there's some ongoing discussion in Slack. I'll leave it up for a bit longer. To summarize so far: 

Why move to markdown in the repo: 
- we frequently introduce breaking APIs in the software. It is useful to the user be able to link the correct version of the documentation with the version of the binary. Git gives that to us for free.
- building on the above, we can also use static site generation software to render the documentation in a more user friendly way (mkdocs, readthedocs, spinx, hugo). 

Why leave it in the wiki:
The github wiki has a lower barrier of entry for users, than cloning the repo and opening a PR to make a docs update. This was the initial reason I did create the wiki, and this model has moderate success for other projects in the community (munki, autopkg). 
Over time, I find the arguments for this model less appealing. Most of the documentation provided to the project is still written by the code contributors, not the users. IMO shifting away from the wiki, with the goal to get the documentation viewable both in the repo and on micromdm.io would have more benefit to end users and the project. 

@GavinElder thanks again for the initiative/effort to convert this. I will leave to @jessepeterson to decide and merge. If we go down this route we should consider removing the wiki pages and redirecting links to the git repo.

gavinelder:
Closing as it's became stale.

PR #563, opened 2019-01-29
Vpp platform

jenjac:
Connects the VPP package with Platform and starts with endpoints to get licenses and service config information from stored VPP Tokens.

### DIFF
--- a/platform/vpp/client.go
+++ b/platform/vpp/client.go
@@ -30,8 +30,8 @@ func NewHTTPClient(instance, token string, logger log.Logger, opts ...httptransp
 	var getVPPServiceConfigSrvEndpoint endpoint.Endpoint
 	{
 		getVPPServiceConfigSrvEndpoint = httptransport.NewClient(
-			"GET",
-			httputil.CopyURL(u, "/v1/vpp/vppserviceconfigsrv"),
+			"POST",
+			httputil.CopyURL(u, "/v1/vpp/serviceconfigsrv"),
 			httputil.EncodeRequestWithToken(token, httptransport.EncodeJSONRequest),
 			decodeGetVPPServiceConfigSrvResponse,
 			opts...,
@@ -39,7 +39,7 @@ func NewHTTPClient(instance, token string, logger log.Logger, opts ...httptransp
 	}
 
 	return Endpoints{
-		GetLicensesSrvEndpoint:    		  getLicensesSrvEndpoint,
-		GetVPPServiceConfigSrvEndpoint:   getVPPServiceConfigSrvEndpoint,
+		GetLicensesSrvEndpoint:         getLicensesSrvEndpoint,
+		GetVPPServiceConfigSrvEndpoint: getVPPServiceConfigSrvEndpoint,
 	}, nil
 }
--- a/platform/vpp/client.go
+++ b/platform/vpp/client.go
@@ -1,0 +1,45 @@
+package vpp
+
+import (
+	"net/url"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/log"
+	httptransport "github.com/go-kit/kit/transport/http"
+
+	"github.com/micromdm/micromdm/pkg/httputil"
+)
+
+func NewHTTPClient(instance, token string, logger log.Logger, opts ...httptransport.ClientOption) (Service, error) {
+	u, err := url.Parse(instance)
+	if err != nil {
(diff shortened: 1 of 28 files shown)


jenjac:
Just added mdmctl commands to interact with a VPP token

jenjac:
This Pull Request now includes everything that you should need to add a single VPP token, and to view apps and licenses within that token, both through api and through mdmctl get vpp-apps/vpp-licenses.

It can also handle VPP Device Assignment using the api, and mdmctl apply/remove vpp-apps.

cmd/mdmctl/apply.go
```diff
@@ -60,6 +60,8 @@ func (cmd *applyCommand) Run(args []string) error {
 		run = cmd.applyProfile
 	case "app":
 		run = cmd.applyApp
+	case "vpp-apps", "vpp-app":
```
GavinElder: why both ?

vpp/managevpplicensesbyadamidsrv.go
```diff
@@ -1,86 +1,91 @@
...
-	response, err := c.ManageVPPLicensesByAdamIdSrv(appID, options)
-	return &response, err
-}
-
-// Interfaces with the ManageVPPLicensesByAdamIdSrv to managed VPP licenses
-func (c *Client) ManageVPPLicensesByAdamIdSrv(appID string, options ManageVPPLicensesByAdamIdSrvOptions) (ManageVPPLicensesByAdamIdSrv, error) {
-	options.SToken = c.SToken
-	options.AdamIDStr = appID
+	if options.AdamIDStr == "" {
+		return nil, errors.Wrap(nil, "must include adamIdStr")
```
GavinElder: `AdamIDStr` ?

platform/vpp/client.go
```diff
@@ -0,0 +1,93 @@
...
+			httputil.CopyURL(u, "/v1/vpp/licenses"),
+			httputil.EncodeRequestWithToken(token, httptransport.EncodeJSONRequest),
+			decodeGetLicensesSrvResponse,
+			opts...,
+		).Endpoint()
+	}
+
+	var manageVPPLicensesByAdamIdSrvEndpoint endpoint.Endpoint
+	{
+		manageVPPLicensesByAdamIdSrvEndpoint = httptransport.NewClient(
```
GavinElder: nit, seems that a number of the vars swap between `Id` and `ID` I believe it should be consistently applied as ID as per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

icanhazpython:
Hi all, what happened to this PR? What is the current status? VPP support would be great, and it seems like most or all of the code is already written and waiting to be merged. It's a damn shame it's sitting here rotting.

hntrl:
bump

brevilo:
Same here. After dropped its own free MDM solution (Apple Server) we're looking for an alternative. VPP/ABM is the *only* part we actually use (just license management, without any DEP). So seeing this merged would solve our problems. What's left/needed to get this PR merged?

krmzbbr:
I was able to install VPP custom app using these steps:

You need to assign licenses to serial numbers using POST request to the VPP management endpoint (with Bearer token you will get from Apple School/Business manager)

`
POST
https://vpp.itunes.apple.com/mdm/v2/assets/associate
{
    "assets": [
        {
            "adamId": "99999999",
            "pricingParam": "STDQ"
        }       
    ],    
    "serialNumbers": [
        "DMQZXXXXXXXX",
    ]
}
`
This will let you install VPP apps but I had to use a custom script instead of install_appstore_application because it requires you to submit itunes_store_id (as number) instead of an identifier. If you use identifier, you will get a "Could not retrieve license for the app" from MDM - I have wasted so many hours until I figured this out. 

`
#!/bin/bash  
source $MICROMDM_ENV_PATH
endpoint="v1/commands"
jq -n \
  --arg request_type "InstallApplication" \
  --arg udid "$1" \
  --arg itunes_store_id $2 \
  '.udid = $udid
  |.request_type = $request_type
  |.itunes_store_id = ($itunes_store_id|tonumber)
  |.options = {"purchase_method": 1}
  '|\
  curl $CURL_OPTS \
    -H "Content-Type: application/json" \
    -K <(cat <<< "-u micromdm:$API_TOKEN") "$SERVER_URL/$endpoint" -d@-
`

GregOriol:
I can confirm that the process described by @krmzbbr works

PR #577, opened 2019-02-27
variable scep client validity

bdemetris:
allows admins to set a desired scep client validity in days as an argument to `micromdm serve`.  default will still be 365.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Remove deprecated `-apns` flags from server startup (#528)
 * Move API calls to list endpoints from HTTP GET to HTTP POST (#522, #523, #524, #525, #526)
 * Add support for the ApplicationConfiguration Setting (#521)
+* Allow SCEP client validity to be adjusted via server startup flag (#577)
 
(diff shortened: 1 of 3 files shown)


jessepeterson:
I like it! I'd suggest changing the switch name to `scep-client-expiry` (or similar) to make it super clear, and I'd also update the `CHANGELOG.md` to add a reference to your change.

bdemetris:
updated the changelog as requested.  I am not sure I want to change the flag name here.  The functions for scep client validity are called as such, so having a server flag that doesn't reflect the variable names would be possibly even more confusing.  Also, we aren't measuring an expiry in this case since that would be defined as a target date (date of expiration), we are defining how long the certificate is valid for (validity, or valid for length in days), so i'd like to stick with validity :)

jessepeterson:
Validity is fine, but it has no real bearing on the colloquial 'validity' of device identity certificates. Expired ones are completely valid for MicroMDM — and work just fine. I agree expiry _can be thought of_ as a target date, but doesn't need to be. The idea behind hinting at the expiration is this is what people are _actually_ trying to set — and so it seems more 'discoverable' from a UX perspective. But, yes, in the X.509 world 'validity' is commonly used to refer to the expiration period. I don't feel that strongly about it. :)

PR #578, opened 2019-03-01
Handle ActivationLockBypassCode Request

jenjac:
Allow a request to be sent to an iOS device to get an activation lock bypass code.

### DIFF
--- a/mdm/mdm/command.go
+++ b/mdm/mdm/command.go
@@ -60,7 +60,10 @@ type Command struct {
 	ScheduleOSUpdateScan            *ScheduleOSUpdateScan
 	ActiveNSExtensions              *ActiveNSExtensions
 	RotateFileVaultKey              *RotateFileVaultKey
+	ActivationLockBypassCode        *ActivationLockBypassCode
 }
+
+type ActivationLockBypassCode struct{}
 
 // InstallProfile is an InstallProfile MDM Command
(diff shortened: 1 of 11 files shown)


knightsc:
Couple quick comments. Can you add an entry to the CHANGELOG for this? Also could you add a script example of sending this command in the /tools/api/commands folder?

jenjac:
Updated CHANGELOG and added api script

mdm/mdm/command.go
```diff
@@ -60,8 +60,11 @@ type Command struct {
 	ScheduleOSUpdateScan            *ScheduleOSUpdateScan
 	ActiveNSExtensions              *ActiveNSExtensions
 	RotateFileVaultKey              *RotateFileVaultKey
+	ActivationLockBypassCode        *ActivationLockBypassCode
 }
 
+type ActivationLockBypassCode struct{}
```
knightsc: https://macadmins.slack.com/archives/C19RTE0L9/p1551718526006700

See the conversation in MacAdmins slack. I don't believe we need the explicit type here or in the list above. Since the command takes no params. I think you can remove the changes in this file but keep the marhsalling code and the test command should still work

jenjac: I missed that one when I simplified the code. I updated it now.

Issue #580, opened 2019-03-21
set the time when moving an ack'd command to a completed queue


groob:
https://github.com/micromdm/micromdm/blob/8e0d722d61cfd50b631f15912488f9ac2de3117e/platform/queue/queue.go#L75-L82

When a command is ack'd the command's Acknowledged time needs to be set.

bdemetris:
this actually extends to the other queues as well.  `time` handles this as "zero time" and outputs `1754-08-30 22:43:41.128654848 +0000 UTC` to all items in the command struct

PR #581, opened 2019-03-21
set acknowledge time when moving command to completed queue

bdemetris:
quick resolve for https://github.com/micromdm/micromdm/issues/580

this just ensures that each time a command is moved to completed it's time is recorded in the struct.  useful to prune completed commands by date.

### DIFF
--- a/platform/queue/queue.go
+++ b/platform/queue/queue.go
@@ -4,6 +4,7 @@ package queue
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/boltdb/bolt"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


platform/queue/queue.go
```diff
@@ -75,6 +76,7 @@ func (db *Store) nextCommand(ctx context.Context, resp mdm.Response) (*Command,
 	case "Acknowledged":
 		// move to completed, send next
 		x, a := cut(dc.Commands, resp.CommandUUID)
+		x.Acknowledged = time.Now()
```
mariozig: Thoughts on renaming this to `AcknowledgedAt` here and in device_command.go?  I'm thinking it could make someone reading this at a call site understand it's a timestamp rather than a boolean.  (also other time properties are suffixed with `At`)

groob: That's not a bad idea, but it should be a separate issue/PR. please file it.

groob: Use `time.Now().UTC()`

PR #589, opened 2019-05-29
Allow for getting full device data

jenjac:
Allow returning all device data through the api by returning an array of Device objects rather than just a subset using the DeviceDTO object.

### DIFF
--- a/cmd/mdmctl/get.go
+++ b/cmd/mdmctl/get.go
@@ -145,7 +145,7 @@ func (cmd *getCommand) getDevices(args []string) error {
 		return err
 	}
 	for _, d := range devices {
-		fmt.Fprintf(out.w, "%s\t%s\t%v\t%s\n", d.UDID, d.SerialNumber, d.EnrollmentStatus, d.LastSeen)
+		fmt.Fprintf(out.w, "%s\t%s\t%v\t%s\n", d.UDID, d.SerialNumber, d.Enrolled, d.LastSeen)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


groob:
I’m skeptical this is a good idea. My intention is to actually trim the fields we save to an absolute minimum. 

I would like to push off any inventory / host state type of data to be built in a higher level project which uses micromdm as a dependency.

jessepeterson:
I concur with @groob here. This is exactly what the API is for.

jenjac:
What I specifically need in my project is to be able to get the ProductName from a device so I can tell which type of device it is, even if it is not enrolled in DEP.  I also need to be able to get the UnlockToken.  The information is there, but there is not currently any way to access it.

jenjac:
Where in the API can you access that information, if it isn't sent with the device information?

groob:
You can get ProductName by issuing a DeviceInformation query, no?

jenjac:
Yes, but since it is already being stored in the micromdm db, I was hoping I would be able to get it from there, instead of storing a duplicate myself, or issuing a query and waiting for a response every time.

groob:
The things that micromdm keeps in its database are there only to make micromdm be able to function. 
As I mentioned in my initial post, if you want to maintain any inventory specific data, you should use the  DeviceInformation query, and subscribe to the webhooks to build inventory data in your own database. 

The main reason I have this opinion is that once you're relying on micromdm to tell you this data, it opens the door for making this data reliable, and micromdm has to decide to implicitly run queries at an interval to track state. It also forces us to accept more features that expand the API.

groob:
However, if you can make a case that a specific field can only be obtained by MicroMDM storing that information, we should expose that in the `DeviceDTO` structure.

PR #592, opened 2019-06-08
Serialization of InstallApplicationOptions is fixed.

nkllkc:
There was a problem of serialization for InstallApplicationOptions.PurchaseMethod = 0.


groob:
Also see how current zeroes are handled here: 
https://github.com/micromdm/micromdm/blob/577e4c9538226313643f93e55ae28673eb2b0cf3/mdm/mdm/marshal_proto.go#L179

A deeper refactor might be required to account for a desired zero value.

PR #597, opened 2019-06-29
Support User Enrollment

groob:
In iOS 10.13/macOS 10.15 a new, BYOD specific enrollment type was added, called User Enrollment. 
This enrollment type replaces the typical `UDID` field in checkin and acknowledge requests with a `EnrollmentID` field which is unique per each enrollment. One important aspect of this enrollment type is that no personally identifiable information is available to the MDM (UDID, SerialNumber). 

The implementation implemented here adds the new EnrollmentID field where appropriate, and ensures that the device tables do not store the enrollment ID. 

I will follow up this change set with one that allows listing/removing current enrollment IDs in a similar way that `mdmctl get devices` and `mdmctl get users` does.


btctcn:
@groob Hi! I just enroll my iPhone in User Enrollment mode. But `./tools/api/get_devices` command displays only devices enrolled in Device Enrollment mode. Are there any plans to improve the command to support devices enrolled in User Enrollment mode. (v.1.7.1).

groob:
Pull requests welcome :) 

At least, consider opening a new issue with this specific feature request.

Issue #456, opened 2018-07-12
mdmctl needs server.json before mdmcert is used


jessepeterson:

jessepeterson:
Perhaps it goes without saying but, as a workaround, you can defeat this by either placing an empty JSON server config in `~/.micromdm/servers.json` (e.g. `{}`) or by simply doing the `mdmctl config` [steps](https://github.com/micromdm/micromdm/wiki/Quickstart#connecting-mdmctl).

PR #613, opened 2019-09-04
allowed csr creation without a config file; added min macOS version f…

tperfitt:
when integrating micromdm into the MDS app, i found a couple of things that would make integration with apps easier:

1. The build file should specify a minimum macOS build number or else notarization will fail since the MachO binary won't have that value.

2. Generating a CSR request doesn't require the server component expect for logging. I wanted to use the CSR generation in mdmctl, but don't want to create a configuration file. So I added in a nolog flag that doesn't require talking to the server. To do this, I had to move setup for the other operations into each command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,10 @@ else
 endif
 
 export GO111MODULE=on
-CGO_CFLAGS=-mmacosx-version-min=10.12 
+
+#Specify a minimum version for macos otherwise notarization will fail
 CGO_LDFLAGS=-mmacosx-version-min=10.12 
+
 VERSION = $(shell git describe --tags --always --dirty)
 BRANCH = $(shell git rev-parse --abbrev-ref HEAD)
 REVISION = $(shell git rev-parse HEAD)
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,8 @@ else
 endif
 
 export GO111MODULE=on
-
+CGO_CFLAGS=-mmacosx-version-min=10.12 
+CGO_LDFLAGS=-mmacosx-version-min=10.12 
 VERSION = $(shell git describe --tags --always --dirty)
 BRANCH = $(shell git rev-parse --abbrev-ref HEAD)
 REVISION = $(shell git rev-parse HEAD)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -9,7 +9,8 @@ else
 endif
 
 export GO111MODULE=on
-
+CGO_CFLAGS=-mmacosx-version-min=10.12
```
groob: Add a comment or I'll forget why this is here.

cmd/mdmctl/mdmcert.go
```diff
@@ -102,6 +98,11 @@ const (
 )
 
 func (cmd *mdmcertCommand) runVendor(args []string) error {
+    
+    if err := cmd.setup(); err != nil {
```
groob: gofmt

cmd/mdmctl/mdmcert.go
```diff
@@ -169,6 +170,7 @@ func (cmd *mdmcertCommand) runPush(args []string) error {
 		flCN       = flagset.String("cn", "micromdm-user", "CommonName for the CSR Subject.")
 		flPKeyPass = flagset.String("password", "", "Password to encrypt/read the RSA key.")
 		flKeyPath  = flagset.String("private-key", filepath.Join(mdmcertdir, pushCertificatePrivateKeyFilename), "Path to the push certificate private key. A new RSA key will be created at this path.")
+		flNoLogging	 = flagset.Bool("nolog",false,"Does not log request to server so not configuration file required.")
```
groob: `log` is an odd choice for sending an RPC. `-local-only` or `-no-upload` would be more clear.

jessepeterson:
Issue also reported in #456

groob:
@tperfitt you need to run `gofmt` (usually helpful to have as an autosave option for your editor)

https://blog.golang.org/go-fmt-your-code

tperfitt:
I ran gofmt on it and updated the pull request. Anything else I need to do?

groob:
This is good to merge. I've been thinking about an implementation that doesn't configure the remote options for commands that are local only, which would avoid having an extra flag. 

I'll either merge this or propose a different PR, but I've had no time to do either. Hoping to get some time this week.

PR #617, opened 2019-10-02
Add DEPProfileStatus to API response

grahamgilbert:
Right now this doesn't work - but it should from what I can see. Opening this PR so we can get this working.

This actually returns
```
{
  "devices": [
    {
      "serial_number": "SERIAL",
      "udid": "UDID",
      "enrollment_status": true,
      "last_seen": "2019-09-05T01:32:33.994054Z",
      "dep_profile_status": ""
    }
  ]
}
```

But I would expect it to default to `empty` looking at https://github.com/micromdm/micromdm/blob/master/platform/device/device.go#L45-L53

### DIFF
--- a/platform/device/get_devices.go
+++ b/platform/device/get_devices.go
@@ -19,10 +19,11 @@ type ListDevicesOption struct {
 }
 
 type DeviceDTO struct {
-	SerialNumber     string    `json:"serial_number"`
-	UDID             string    `json:"udid"`
-	EnrollmentStatus bool      `json:"enrollment_status"`
-	LastSeen         time.Time `json:"last_seen"`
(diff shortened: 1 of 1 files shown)


groob:
Any reason not to query DEP directly via the API? you can get this information by calling `/v1/dep/devices`, which is the API `mdmctl` itself uses for `mdmctl get dep-devices -serials=(serial)` 

I initially had this larger devices structure, but later decided it should be kept as small as possible, so that micromdm itself doesn't have to keep a consistent state of various device fields.

groob:
On the other hand this is already synced internally, so exposing it will just work. https://github.com/micromdm/micromdm/blob/127955dc748c82d4301f02343280631b1bf167a0/platform/device/worker.go#L86

PR #626, opened 2019-11-25
store all device responses

groob:
for science


groob:
Closing but might revisit in a new PR.

PR #633, opened 2020-01-06
Save a record of the last imported DEP token and prefer that one.

### LINKED ISSUE #546: Not able remove DEP-token
1.4.0

mdmctl apply dep-tokens -import ...

I have problem with imported  two DEP-Tokens don't know how to remove or clean imported tokens to get correct one working.

Couse only first token is apply but no more valid in my case.

How can i remove that token.

Thank you

jessepeterson:
The server uses only the very first DEP Token when it starts up, here:

https://github.com/micromdm/micromdm/blob/ea393e9804393beab0da9f4b8d2ae3d251828a68/server/server.go#L263-L272

In order to better facilitate this try to order the returned DEP token array by preferring the last one that was added first.

Fixes #546. _Technically_ this isn't quite a "fix" as it's not a way to delete tokens, but I think fixes the problem that occurs when updating a DEP token.

### DIFF
--- a/platform/config/builtin/db_deptoken.go
+++ b/platform/config/builtin/db_deptoken.go
@@ -22,6 +22,10 @@ func (db *DB) AddToken(consumerKey string, json []byte) error {
 		if err != nil {
 			return err
 		}
+		err = b.Put([]byte("last_added"), []byte(consumerKey))
+		if err != nil {
+			return err
+		}
 		return b.Put([]byte(consumerKey), json)
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jessepeterson:
> Is there a reason we append instead of overriding the token?

@groob I think the idea was that we'd support multiple DEP tokens (and thus DEP clients/syncers).

Technically speaking we store the tokens keyed by the ConsumerKey, so we don't have a "single token" location but rather a bucket of tokens. We'd have to go clear the old one(s), or only store a single token.

Issue #310, opened 2017-11-07
Support secure/limited enrollment


jessepeterson:
In a security-sensitive environment you may not want just any device to be able to enroll. Currently any device that can get a SCEP certificate is able to enroll and have (potentially sensitive) MDM commands sent to it.

A couple strategies for securing this:

* Allow only DEP enrollments for pre-approved (or DEP API-synced) serial numbers
  * I.e. disallow `GET /mdm/enroll` handing over the Profile and only allow the `POST /mdm/enroll` which is both signed by an Apple-only certificate and provides the serial number and a few other attributes. This latter process is also how OTA enrollment is done as well as DEP.
* Gate enrollment profile behind some sort of authentication/approval process (perhaps SAML/OAuth/API/etc.)
* Combine the above techniques with unique and/or time-limited SCEP challenges (or, alternatively, key-like CSR attributes)

groob:
I have a slightly antagonizing proposal here. 

1) remove(disable) the ability for the core server to serve enrollment profiles
2) build a smarter server that just is capable of doing enrollments in a custom way.

I'm sort of trolling here, but also pushing us to think what's in scope for the project to maintain, and what belongs in a workflows service.

omeid:
This is indeed a problem, many enterprise push apps with no little to no security that can access sensitive information.

> Gate enrollment profile behind some sort of authentication/approval process (perhaps SAML/OAuth/API/etc.)

I don't have any exact solution in general, but I would like to point that this solution often ends up in a common account with insecure password shared around organizations which not only doesn't add real security, but worst yet creates the illusion of security.


> I'm sort of trolling here, but also pushing us to think what's in scope for the project to maintain, and what belongs in a workflows service.

I think any decent MDM solution should support _secure_  enrollment :)

Beanow:
Producing acceptable profiles elsewhere is easy so removing that is not sufficient. In our use case we require explicit single use invite codes to be programmatically obtained from another part of our application, not having this code should result in the enrolment being denied.

Beanow:
Possibly #81 related

Issue #81, opened 2017-03-20
Issue unique SCEP challenges per device.


groob:
Right now the embedded SCEP server does not issue any challenge passwords for SCEP. 
MicroMDM should create (and keep track of) a unique password for each enrollment and only allow a single use. 

Alternatively create a short lived token which would expire in <1 minute( but this would cause issues with enrollment profiles which need to be emailed to a user).

jessepeterson:
I think the 'each-enrollment-token' would work very well. As an option an admin can "bless" one of the tokens with an expiry of >1min for re-usable enrollment profile scenarios.

loyaltyarm:
What about in "non-user" scenarios, where devices need to be enrolled but users are arbitrary in the case of kiosk devices. Blessing the token would be the best option here, would the token be set within the enrollment `.mobileconfig`?

groob:
Yes, if you look at what is being returned by /mdm/enroll, a SCEP payload is there, and it should container the scep challenge for that device.

Beanow:
There's a related need we have coming from SimpleMDM, which is including a single use code in the checkin url to know who initiated the enrolment. This allows a kind of invite token metadata to be added to the enrolment process.

WardsParadox:
Did not meant to reference this issue and accidentally close it. Sorry @groob

jessepeterson:
Resolved with #642. Also I think server parameter passing through to the webhook fulfills some of this, too.

PR #642, opened 2020-01-23
dynamic SCEP challenge support

jessepeterson:
This integrates optional dynamic SCEP challenge support. This means that the MicroMDM SCEP server will require a pre-generated random SCEP challenge before issuing a certificate (and thus allowing enrollment into MDM) if the `-use-dynamic-challege` is set. In order for a device to successfully enroll using this method the SCEP challenge must be present in the enrollment profile SCEP payload that is sent to the device. Think of it like an enrollment passcode.

To generate your own enrollment profiles you'll want to use this new API endpoint of `/v1/challenge` which generates, stores, and returns the one-time SCEP challenge which should be embedded in an enrollment profile.

Interestingly, the SCEP server code for this has been written for quite some time (thanks, @groob!). I believe this hadn't been integrated yet because the security value was debatable in that enrollment profiles are automatically generated by MicroMDM for MDM, OTA, and profile enrollments. If profiles (and thus challenges) are simply provided in the profile the the point of even having the challenges is lessened (as there's nothing authenticating the use of a SCEP challenge). However with macOS 10.15 Catalina and iOS 13 we now have a DEP capable way of doing modern authentication and thus proper gating of the enrollment profile (and any sensitive content, like a SCEP payload challenge).

That said this brings up some options of how we want to handle this feature. I see a couple options forward:

1. Require the administrator to completely handle the profile service. For dynamic SCEP challenges to work, generally, something has to dynamically compose enrollment profiles with SCEP challenges embedded in them. This is a lot of logic, presumably on top of an authentication system to handle. It means, basically, standing up another service to handle the enrollment profile delivery. 
2. Implement an optional webbook/callback for the enrollment profile service. I.e. when `/mdm/enroll` is hit, we, essentially, proxy that request to another URL much like # 1 above. This has the benefit that we can still handle some of the lower-level client validation (like PKCS#7 certificate verification on the OTA and DEP enrollment requests) while handing handling the profile generation off to another service.
3. Implement dynamic SCEP challenge profile creation. There's two (minimalist) ways to do this, I see:
   1. For MicroMDM-generated profiles, simply insert the SCEP challenge as we've been doing for non-dynamic SCEP challenges. This means that for anything more involved as an administrator you have to figure out e.g. # 1 above.
   2. Do a simple find-and-replace on any locally generated or stored enrollment profile for a value such as "${SCEP_CHALLENGE}". This has the benefit of working for stored enrollment profiles (i.e. those that an admin has modified) as well as the defaul (generated) profile.

When thinking about this we should make a delineation between "legacy" enrollment endpoints (OTA, "old" DEP, profile endpoint) and new Web view iOS 13/Catalina enrollments. If we optimize for the latter then # 1 seems more appropriate.

Currently, this PR implements 3i above with an optional parameter of `-gen-dynamic-challege` (otherwise, it doesn't mess with the enrollment profile and you're back to situation # 1, above: doing it yourself). I feel like this is a nice compromise of turning it on quickly for experimentation, and having to do everything yourself right off the bat with # 1.

Anyway I'm interested in any feedback!

### DIFF
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -124,20 +124,20 @@ func serve(args []string) error {
 		return errors.Wrapf(err, "creating config directory %s", *flConfigPath)
 	}
 	sm := &server.Server{
-		ConfigPath:        *flConfigPath,
-		ServerPublicURL:   strings.TrimRight(*flServerURL, "/"),
-		Depsim:            *flDepSim,
-		TLSCertPath:       *flTLSCert,
-		CommandWebhookURL: *flCommandWebhookURL,
-		NoCmdHistory:      *flNoCmdHistory,
-		UseDynSCEPChal:    *flUseDynChallenge,
-		GenDynSCEPChal:    *flGenDynChalEnroll,
+		ConfigPath:          *flConfigPath,
+		ServerPublicURL:     strings.TrimRight(*flServerURL, "/"),
+		Depsim:              *flDepSim,
+		TLSCertPath:         *flTLSCert,
+		CommandWebhookURL:   *flCommandWebhookURL,
+		NoCmdHistory:        *flNoCmdHistory,
+		UseDynSCEPChallenge: *flUseDynChallenge,
+		GenDynSCEPChallenge: *flGenDynChalEnroll,
 
 		WebhooksHTTPClient: &http.Client{Timeout: time.Second * 30},
 
 		SCEPClientValidity: *flSCEPClientValidity,
 	}
-	if !sm.UseDynSCEPChal {
+	if !sm.UseDynSCEPChallenge {
 		// TODO: we have a static SCEP challenge password here to prevent
 		// being prompted for the SCEP challenge which happens in a "normal"
 		// (non-DEP) enrollment. While security is not improved it is at least
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -34,6 +34,7 @@ import (
 	appsbuiltin "github.com/micromdm/micromdm/platform/appstore/builtin"
 	"github.com/micromdm/micromdm/platform/blueprint"
 	blueprintbuiltin "github.com/micromdm/micromdm/platform/blueprint/builtin"
+	"github.com/micromdm/micromdm/platform/challenge"
 	"github.com/micromdm/micromdm/platform/command"
 	"github.com/micromdm/micromdm/platform/config"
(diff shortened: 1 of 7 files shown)


jessepeterson:
These topics have been discussed before. See #81, #310, #320. Specifically @groob mentioned (trolled) [outright _removing_ the enrollment profile service](https://github.com/micromdm/micromdm/issues/310#issuecomment-342375610), very similar to the proposal # 1 above.

cmd/micromdm/serve.go
```diff
@@ -127,15 +130,19 @@ func serve(args []string) error {
 		TLSCertPath:       *flTLSCert,
 		CommandWebhookURL: *flCommandWebhookURL,
 		NoCmdHistory:      *flNoCmdHistory,
+		UseDynSCEPChal:    *flUseDynChallenge,
```
groob: Nit: Chal is a odd abbreviation, would prefer the full word personally.

groob: (Applies to all the usages of `Chal` in this PR.

jessepeterson: To be perfectly honest I was annoyed at the larger diff changing the column width of the struct so made it shorter (most diff tools have 'ignore whitespace' default off, like GH. I can update it, heh.

jessepeterson:
> I do feel like there could be a better implementation of the challenge package that you can plug in if you're going to depend on this.

What do you mean?

jessepeterson:
I'm going to merge this since it's approved, but I want to keep the dialog open about best practices here.

To be clear the ideal workflow is that you set `use-dynamic-challege` to true (which then will require valid pre-generated SCEP challenges), but set `gen-dynamic-challege` to **false**. In this way you must get the SCEP challenge from the (authenticated) API and generate the enrollment profile _yourself_ — ideally gated behind some sort of auth/2fac/sso/saml/oauth/etc.

`gen-dynamic-challenge` (the setting that does the dynamic SCEP challenge embedding for you) is basically just a test/troubleshooting setup to make sure the dynamic SCEP challenge setting is working. Or perhaps if you gate the entire `/mdm/enroll` endpoint yourself (in which case allowing MicroMDM to do the generation might be okay). But it's not intended to be used by itself and have direct access to it.

cmd/micromdm/serve.go
groob: Yeah, it's annoying but I'd prefer that over having this variable name for eternity :p

groob:
Merge away!

PR #643, opened 2020-01-23
UDID Cert Auth: warn only option

jessepeterson:
Currently the UDID Cert Auth middleware (#358) rejects connections for mismatched certificates. However, if you have a legitimate reason to allow non-matching certificates (say, you have active enrollments for which [end-users renewed their own certificates](https://github.com/micromdm/micromdm/wiki/Device-Identity-Certificate-Expiration)) you may want to disable the connection rejection while you remediate those enrollments.

I'm not entirely convinced this should even be merged as it actively subverts a security measure. Thoughts welcome.


groob:
Can you clarify what you mean by "warn"? There's no actual warning being printed/displayed anywhere, at least as far as the diff shows.

jessepeterson:
@groob I'm using "warn" here loosely. It already unconditionally logs here:

https://github.com/micromdm/micromdm/blob/127955dc748c82d4301f02343280631b1bf167a0/platform/device/udidauth.go#L53-L60

This merely stops returning errors if it's switched on.

PR #651, opened 2020-03-09
Add EnableRemoteDesktop and DisableRemoteDesktop

tomaswallentinus:
Add ability to send EnableRemoteDesktop and DisableRemoteDesktop mdm-commands:
https://developer.apple.com/documentation/devicemanagement/enable_remote_desktop


tomaswallentinus:
Am I suppose to ad an test in TestEndToEnd as well or is that just for endpoints that have an base64 encoded payload?

PR #677, opened 2020-06-18
add activation lock bypass code creation

groob:
Codes to lock and unlock the device. 

https://developer.apple.com/documentation/devicemanagement/device_assignment/activation_lock_a_device/creating_and_using_bypass_codes

Send the hash to mdmenrollment.apple.com to lock.
To unlock, either type the human readable code string in Setup Assistant, or send it to deviceservices-external.apple.com/deviceservicesworkers/escrowKeyUnlock.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ cmd/micromdm/micromdm
 tools/ngrok/config_root
 tools/ngrok/filerepo
 tools/ngrok/env
+mdm-files/
--- a/dep/activation_lock.go
+++ b/dep/activation_lock.go
@@ -1,0 +1,34 @@
+package dep
+
+import "github.com/pkg/errors"
+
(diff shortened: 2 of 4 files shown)


feishi312:
Hi, how can I use the activation lock bypass code feature? For example, what is the command to call it in bash?

xujimu:
What is the bypass code obtained from here used for? I used the bypass code obtained from here to unlock a device but it failed: https://developer.apple.com/documentation/devicemanagement/get_the_bypass_code_for_activation_lock"

jessepeterson:
Read more about activation unlock here: https://blog.kandji.io/how-to-manage-activation-lock

This is different from clearing the passcode: https://developer.apple.com/documentation/devicemanagement/clear_the_passcode

PR #678, opened 2020-06-22
Updated URL to join MacAdmins Slack group

MobileDan:
The URL to join MacAdmins Slack group was an error page at "macadmins.herokuapp.com". I assume this is the old address and should be "www.macadmins.org".


groob:
Thanks. I just made the same fix on the website!

PR #679, opened 2020-06-24
add auto setup admin accounts changes

groob:
needs some cleanup for the password util, otherwise works great.


jessepeterson (approved):
lgtm

PR #690, opened 2020-07-17
MarshalQueuedCommand property mismatch

MobileDan:
When a QueueCommandQueued is marshaled in MarshalQueuedCommand, the QueueCommandQueued.DeviceUDID was copied to the protobuf CommandUuid property. Now the CommandUUID property is copied to the protobuf CommandUuid property.

### DIFF
--- a/platform/queue/command_queued.go
+++ b/platform/queue/command_queued.go
@@ -18,7 +18,7 @@ func MarshalQueuedCommand(cq *QueueCommandQueued) ([]byte, error) {
 	}
 	return proto.Marshal(&commandqueued.CommandQueued{
 		DeviceUdid:  cq.DeviceUDID,
-		CommandUuid: cq.DeviceUDID,
+		CommandUuid: cq.CommandUUID,
 	})
 }
 


MobileDan:
Hi Jesse,

I am combing through almost every line of code for certain features such as the command queue. When I fix something like this in my branch, I make a note to suggest it as a pull request. I’ve made a few similar suggestions recently.

I noticed this mismatch when I was documenting the workflow of a command coming into MicroMDM through the Web API all the way to when it is executed. I was following where QueueCommandQueued.DeviceUDID was being used by finding all usages in GoLand when I noticed this.

I submitted a pull request only because it bugged me. It does not cause any issues as implemented since CommandUUID is not used but it could cause a problem in the future for this or downstream implementations. I appreciate this project and want to contribute back what I can even if it’s something small.

I’d like to talk with you and Victor about contributing more to the next version. I have created a separate data layer and can easily switch between MySQL, Bolt and a Web API backed datastore. Or, I can intermix them. I still have more work to complete it but plan to contribute back anything I can of value. I’d like to see other’s attempts at this and collaborate. I learned a great deal from the My SQL pull request from Lepidopteron.

 I can also give feedback on what is confusing to newbies. I am an experienced developer but new to Go. I have invested a lot of time into learning Go and understanding MicroMDM.

Thank You,

Dan

Issue #646, opened 2020-02-29
Provide an ability to clear command queue


tperfitt:
# What version of `micromdm` are you using? 
1.6.0

When I removed the com.github.micromdm.micromdm.enroll profile via a MDM command and attempted to re-enroll the mac, it was immediately unenrolled. Removing the mdm.DeviceCommands entries in the db resolved the issue. I believe the remove profile command was queue and sent but never removed since it was not acknowledged.  Deleting the device did not resolve the issue.

Feature Request:
MicroMDM should have a way to clear the queue for any pending commands (or all commands) in order to resolve any issues. Please provide a mechanism both mdmctl and the web API.

jessepeterson:
Addressed with #692.

PR #692, opened 2020-07-19
Clear a UDID's command queue during Authenticate Check-in (enrollment)

### LINKED ISSUE #647: Clear command queue when a device is deleted
# What version of `micromdm` are you using?
1.6.0

When a device is removed from micromdm, any existing commands are not cleared from the queue. This can resulting in unexpected behavior when the device is reenrolled in the future. 

When a device is removed from micromdm, scrub any entries related to that device from the command queue. Future Tim thanks you.

jessepeterson:
Fixes #647 and addresses the mentioned symptom in #646.

### DIFF
--- a/mdm/checkin.go
+++ b/mdm/checkin.go
@@ -27,6 +27,13 @@ func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) error {
 		return errors.Wrap(err, "get checkin topic from message")
 	}
 
+	if topic == AuthenticateTopic {
+		err = svc.queue.Clear(ctx, event)
+		if err != nil {
+			return errors.Wrap(err, "clearing queue on enrollment attempt")
+		}
+	}
+
 	err = svc.pub.Publish(ctx, topic, msg)
 	return errors.Wrapf(err, "publish checkin on topic: %s", topic)
 }
--- a/mdm/checkin.go
+++ b/mdm/checkin.go
@@ -28,8 +28,7 @@ func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) error {
 	}
 
 	if topic == AuthenticateTopic {
-		err = svc.queue.Clear(ctx, event)
-		if err != nil {
+		if err := svc.queue.Clear(ctx, event); err != nil {
 			return errors.Wrap(err, "clearing queue on enrollment attempt")
 		}
 	}
--- a/platform/queue/queue.go
+++ b/platform/queue/queue.go
@@ -64,22 +64,16 @@ func (db *Store) Clear(ctx context.Context, event mdm.CheckinEvent) error {
 	}
 
 	dc, err := db.DeviceCommand(udid)
-	if err != nil {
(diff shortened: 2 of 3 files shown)


mdm/checkin.go
```diff
@@ -27,6 +27,13 @@ func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) error {
 		return errors.Wrap(err, "get checkin topic from message")
 	}
 
+	if topic == AuthenticateTopic {
```
groob, line 30: Probably doesn't make much of a difference, but `TokenUpdate` is when you really know if an enrollment happened.

mdm/checkin.go
```diff
@@ -27,6 +27,13 @@ func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) error {
...
+		err = svc.queue.Clear(ctx, event)
```
groob, line 31: prefer `if err := `

platform/queue/queue.go
```diff
@@ -54,6 +54,34 @@ func (db *Store) Next(ctx context.Context, resp mdm.Response) ([]byte, error) {
...
+	udid := event.Command.UDID
+	if event.Command.UserID != "" {
+		udid = event.Command.UserID
+	}
+	if event.Command.EnrollmentID != "" {
+		udid = event.Command.EnrollmentID
+	}
+
+	dc, err := db.DeviceCommand(udid)
+	if err != nil {
```
groob, line 67:
```
if isNotFound { return nil }
else if err != nil { // wrap and return }
```

platform/queue/queue.go
```diff
@@ -54,6 +54,34 @@ func (db *Store) Next(ctx context.Context, resp mdm.Response) ([]byte, error) {
...
+		if isNotFound(err) {
+			return nil
+		}
+		return errors.Wrapf(err, "get device command from queue, udid: %s", udid)
```
groob, line 71: Maybe `get device to clear queue`?

platform/queue/queue.go
```diff
@@ -54,6 +54,34 @@ func (db *Store) Next(ctx context.Context, resp mdm.Response) ([]byte, error) {
...
+	}
+
+	dc.Commands = nil
+	dc.NotNow = nil
+
+	err = db.Save(dc)
+	if err != nil {
+		return err
```
groob, line 79: either wrap, or `return db.Save(dc)`

mdm/checkin.go
jessepeterson: This is true. However since this service doesn't know which `TokenUpdate` has happened (the first, or any subsequent) it could potentially clear the queue in the midst of normal MDM operation if we checked on token update. `Authenticate` only happens during initial enrollment.

Separately I considered clearing on a `Checkout` event but decided against it for forensics considerations. Also separately I considered appending the queued commands to the "Failed" queue for forensics but decided that'd be far-fetched.

PR #700, opened 2020-08-13
Option to validate only CA and Issuer rather than entire SCEP cert

grahamgilbert:
Currently we're only verifying whether the cert presented is present in the database. For reasons I don't understand / have time to investigate, clients have valid SCEP certificates which aren't stored in the database. This PR adds an option to validate only the Issuer and the CA that signed the cert.

### DIFF
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -88,7 +88,6 @@ func serve(args []string) error {
 		flUseDynChallenge    = flagset.Bool("use-dynamic-challenge", env.Bool("MICROMDM_USE_DYNAMIC_CHALLENGE", false), "require dynamic SCEP challenges")
 		flGenDynChalEnroll   = flagset.Bool("gen-dynamic-challenge", env.Bool("MICROMDM_GEN_DYNAMIC_CHALLENGE", false), "generate dynamic SCEP challenges in enrollment profile (built-in only)")
 		flValidateSCEPIssuer = flagset.Bool("validate-scep-issuer", env.Bool("MICROMDM_VALIDATE_SCEP_ISSUER", false), "validate only the issuer of the SCEP certificate rather than the whole certificate")
-		flSCEPIssuer         = flagset.String("scep-issuer", env.String("MICROMDM_SCEP_ISSUER", "OU=MICROMDM SCEP CA,O=MicroMDM,C=US"), "the issuer to validate against")
 		flPrintArgs          = flagset.Bool("print-flags", false, "Print all flags and their values")
 	)
 	flagset.Usage = usageFor(flagset, "micromdm serve [flags]")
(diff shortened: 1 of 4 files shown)


server/devicecert.go
```diff
@@ -58,9 +113,20 @@ func (mw *verifyCertificateMiddleware) Checkin(ctx context.Context, req mdm.Chec
 		return errors.Wrap(err, "error checking device certificate")
 	}
 	if !hasCN {
```
groob, line 115: same comments about flow here as in the other verify function.

cmd/micromdm/serve.go
```diff
@@ -87,6 +87,8 @@ func serve(args []string) error {
 		flNoCmdHistory       = flagset.Bool("no-command-history", env.Bool("MICROMDM_NO_COMMAND_HISTORY", false), "disables saving of command history")
 		flUseDynChallenge    = flagset.Bool("use-dynamic-challenge", env.Bool("MICROMDM_USE_DYNAMIC_CHALLENGE", false), "require dynamic SCEP challenges")
 		flGenDynChalEnroll   = flagset.Bool("gen-dynamic-challenge", env.Bool("MICROMDM_GEN_DYNAMIC_CHALLENGE", false), "generate dynamic SCEP challenges in enrollment profile (built-in only)")
+		flValidateSCEPIssuer = flagset.Bool("validate-scep-issuer", env.Bool("MICROMDM_VALIDATE_SCEP_ISSUER", false), "validate only the issuer of the SCEP certificate rather than the whole certificate")
+		flSCEPIssuer         = flagset.String("scep-issuer", env.String("MICROMDM_SCEP_ISSUER", "OU=MICROMDM SCEP CA,O=MicroMDM,C=US"), "the issuer to validate against")
```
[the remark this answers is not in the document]
grahamgilbert: Six of one, half dozen of the other. This way lets us have the default, which I expect most are using.

server/devicecert.go
```diff
@@ -3,32 +3,76 @@ package server
...
+	validateSCEPIssuer bool
+	scepIssuer         string
+	scepDepot          *boltdepot.Depot
+}
+
+func verifyIssuer(devcert *x509.Certificate, scepIssuer string, scepDepot *boltdepot.Depot) (bool, error) {
+	issuer := devcert.Issuer.String()
+	expiration := devcert.NotAfter
+
+	if time.Now().After(expiration) {
```
jessepeterson, line 47: I would either add an option to not validate the device certificate expiry or don't do it by default. This will, effectively, "lock out" devices from being remediated by re-enrollment if they ever expire and, by default, MicroMDM has a 1yr expiration on the device cert — this could catch many folks off-guard and be in an unfixable state.

grahamgilbert:
Changelog updated, added a flag to opt in to checking if the cert is expired.

PR #705, opened 2020-09-11
fixed regression with local-only flag

tperfitt:
the local-only flag was added to mdmctl so that local configuration is not needed for generating a CSR. Contacting the server was only used for logging. There appears to be a regression where logging happens earlier and the local-only flag no longer works. This pull enables logging earlier only if the local-only flag is not set.

### DIFF
--- a/cmd/mdmctl/mdmcert.go
+++ b/cmd/mdmctl/mdmcert.go
@@ -71,9 +71,18 @@ func (cmd *mdmcertCommand) Run(args []string) error {
 		cmd.Usage()
 		os.Exit(1)
 	}
+	
+	localonly := false
+	for _, a := range args {
+		if a == "-local-only" {
+			localonly=true
+		}
+	}
 
-	if err := cmd.setup(); err != nil {
-		return err
+	if !localonly {
+		if err := cmd.setup(); err != nil {
+			return err
+		}
 	}
 
 	var run func([]string) error


groob:
How can I reproduce the bug?

tperfitt:
remove the json config, then request a push certificate using mdmctl and these flags:

    NSArray *arguments=@[mdmctlPath,@"mdmcert",@"push",@"-local-only", @"-country",countryCode,@"-email",email,@"-password",password,@"-out",csrPath,@"-private-key",privateKeyPath];

groob:
Did it ever work? You made the last commit on the file by adding the flag, so I'm not sure I understand what part of this is a regression. 

Looking at the implementation, the `-local-only` flag is added to the `push` subcommand, but the config loading happens before these flags are parsed. If my understanding is correct, the flag should be removed and the implementation should be refactored to be cleaner, rather than adding this patch.

tperfitt:
the config loading was added after i added the flag, so i have no idea why it was added so i was hesitant to refactor without understanding why it was added. i'll look at blame and see who added that.

tperfitt:
ok, no idea how this ever worked since my change (which i had been using) did work and now it doesn't.  I'll refactor.

jessepeterson (changes requested):
Requesting changes tperfitt suggested.

jessepeterson:
Just a ping on this fella @tperfitt :)

PR #711, opened 2020-09-24
Avoid queue save when not needed

jessepeterson:
When there's no new command and there's an `Idle` connect there's no need to re-write the queue for a device back out — nothing's changed. Saves a BoltDB write operation in that case.

### DIFF
--- a/platform/queue/queue.go
+++ b/platform/queue/queue.go
@@ -164,6 +164,8 @@ func (db *Store) nextCommand(ctx context.Context, resp mdm.Response) (*Command, 
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


platform/queue/queue.go
```diff
@@ -164,8 +164,10 @@ func (db *Store) nextCommand(ctx context.Context, resp mdm.Response) (*Command,
 		}
 	}
 
-	if err := db.Save(dc); err != nil {
-		return nil, err
+	if resp.Status != "Idle" || cmd != nil {
```
groob, line 167: can you add a comment above explaining the conditional?

groob (approved):

PR #719, opened 2020-12-16
enabling timezone setting

choehn-signogy:
This change enables micromdm to set timezone on iOS/iPadOS/tvOS >= 14.0

SettingsCommand.Command.Settings.TimeZone
https://developer.apple.com/documentation/devicemanagement/settingscommand/command/settings/timezone

The feature was presented during WWDC20


groob (approved):

ayush5harma:
can someone share a sample usage command

PR #721, opened 2021-01-14
Spruce up built in webpage

williamtheaker:
Used CSS from https://micromdm.io

### DIFF
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -54,15 +54,46 @@ const homePage = `<!doctype html>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<title>MicroMDM</title>
-	<style>
-		body {
-			font-family: -apple-system, BlinkMacSystemFont, sans-serif;
-		}
-	</style>
-</head>
+  <style>
+@font-face {
+  font-family: system;
+  font-weight: 900;
+  src: local(".SFNSText-Heavy"), local(".HelveticaNeueDeskInterface-Heavy"), local(".LucidaGrandeUI"), local("Ubuntu Heavy"), local("Segoe UI Heavy"), local("Roboto-Heavy"), local("DroidSans"), local("Tahoma");
+}
+
+/* Global */
+body {
+  position: relative;
+  font-family: "system", "Helvetica Neue", "Helvetica", "Arial", sans-serif;
+  margin: 0;
+  text-align: center;
+  font-size: 48px;
+}
+
+.enrollment {
+  background: #4a9dff;
+  color: #fff;
+  padding: 15px;
+  border-radius: 3px;
+  font-size: 30px;
+  text-decoration: none;
+}
+  </style>
+ </head>
 <body>
-	<h3>Welcome to MicroMDM!</h3>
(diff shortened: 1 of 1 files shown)


williamtheaker:
Screen Shot 2021-02-05 at 4 24 24 PM

jessepeterson:
Just curious: what does this look like on iOS & iPad? Just want to make sure it's usable there, too.

williamtheaker:
@rylandharris was kind enough to test on iOS devices for me.
iPad 7th generation running 14.4
iPhone XR running 14.3

image0
Image from iOS

groob:
Up to @jessepeterson if we want to merge this in. 

I'm personally not in favor, even though it looks nice. I've been encouraging people to write their own enrollment logic and not use the built in one.

williamtheaker:
I think setting the `homepage` flag to false by default would be a better way to accomplish that than rejecting visual improvements.

jessepeterson:
I like the idea of the nicer home page in general. But to discourage built-in enrollment use, though, what do folks think of these options:

1. Do nothing, merge as is.

2. Add a new flag, disabled by default, that disables `/mdm/enroll` endpoint (and/or turns on a 'helpful' message like "Enrollment disabled" or something.

3. Outright remove the 'Enroll a device' button/link from the *home page* (not the service itself).

4. Disable `-homepage` flag by default as @williamtheaker suggested.

I don't feel super strong about this but I'm leaning toward doing 1 and then creating a new issue for 2/3. I don't really like 4. The homepage is a nice canary when first starting up MicroMDM to tell that its working and getting a 404 right off the bat could be non-intuitive for beginners. Not that MicroMDM is all that beginner friendly, heh. :shrug:

discentem:
> Do nothing, merge as is.
I'd be okay with this. 
> Add a new flag, disabled by default, that disables /mdm/enroll endpoint (and/or turns on a 'helpful' message like "Enrollment disabled" or something.
This is seems like a good option. Though, to be honest, I'd be surprised if there are micromdm users out there who don't rely on the enroll webpage as a fallback. 
> Outright remove the 'Enroll a device' button/link from the home page (not the service itself).
Personally, I'd like to retain this option. So I'd prefer the previously mentioned flag. 
> Outright remove the 'Enroll a device' button/link from the home page (not the service itself).
I'm also okay with this, because it retains the option to use it.

jessepeterson (approved):

jessepeterson (changes requested):
I know it's a tiny page, but could we include some basic accessibility tags/attributes for the SVG? E.g. a title, description, etc. I just googled a few things seems like they'd be easy additions.

jessepeterson:
Created issue #741 for limiting default enrollment capability.

williamtheaker:
> I know it's a tiny page, but could we include some basic accessibility tags/attributes for the SVG? E.g. a title, description, etc. I just googled a few things seems like they'd be easy additions.

Good call out! I didn't take that into account at all. I ran the question by accessibility consultant @bpmcneilly and was told "it should be null - assuming MicroMDM is just gonna be the alt text anyway, and that's displayed."

jessepeterson (approved):
Looks good. Just update the CHANGELOG and we're good to!

PR #724, opened 2021-02-08
replace directive isn't supported by go get except for top level module

discentem:
When I try to import `github.com/micromdm/micromdm/platform/queue` from my own module (vs github.com/micromdm/micromdm) it fails because replace directive isn't respected.

```
go: finding module for package github.com/micromdm/micromdm/platform/queue
go: found github.com/micromdm/micromdm/platform/queue in github.com/micromdm/micromdm v1.7.1
go: github.com/micromdm/micromdm@v1.7.1 requires
	github.com/fullsailor/pkcs7@v0.0.0-20180824154052-36585635cb64: invalid version: unknown revision 36585635cb64
```

Seems like this is WAI: https://github.com/golang/go/issues/30354

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ require (
 	github.com/RobotsAndPencils/buford v0.12.0
 	github.com/boltdb/bolt v1.3.1
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/fullsailor/pkcs7 v0.0.0-20180824154052-36585635cb64
+	github.com/groob/pkcs7 v0.0.0-20180824154052-36585635cb64
 	github.com/garyburd/go-oauth v0.0.0-20180319155456-bca2e7f09a17
 	github.com/go-kit/kit v0.7.0
 	github.com/go-sql-driver/mysql v1.4.0 // indirect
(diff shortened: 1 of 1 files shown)


jessepeterson:
Does this fix your issue? I feel like this would cause an issue with the project itself as the import path would then be wrong (being `groob/pkcs7` instead of `fullsailor/pkcs7`). All that being said we ought to switch out for the Mozilla project was done for scep repo.

discentem:
That's a great callout @jessepeterson, thanks for catching. I did not consider how I might break micro and did not actually test. I don't know if it will fix my issue unless I change the import paths on my fork? Let me try that and get back to you.

groob:
@discentem why don't you add the same replace directive to your own go.mod? Or is that causing a different issue for you?

discentem:
@groob Using my own replace directive solved the issue. Should I leave this open for any reason?

groob:
No, I think that's WAI. When you create a new project in Go you get to pick which dependencies to pin to specific versions. 

Also as Jesse mentioned, it would be best to switch both this project and the scep repo to use the mozilla fork of the pkcs7 library, as that's the best maintained one out there.

PR #744, opened 2021-04-23
add support for proxy headers

korylprince:
Right now, if you run micromdm behind a reverse proxy, logs always show the same request host (the proxy itself). This adds a [middleware](https://pkg.go.dev/github.com/gorilla/handlers#ProxyHeaders) that parses X-Forwarded-For, X-Real-IP, etc headers that reverse proxies set so the downstream handlers (i.e. loggers) see the real client IP.

This is hidden behind a default-off flag, as you don't want to parse these headers unless you're behind a trusted proxy.


jessepeterson:
If this is just for logging I might suggest this happens in groob/finalizer. Reason being that because we're using structured logs we can just as well log both the remote addr and the rproxy headers together. No need to lose the actual remote ip (say, in the case you have multiple proxies).

But I dunno, I could see it both ways. Thoughts?

korylprince:
The middleware only modifies the `request.RemoteAddr`, `request.URL.Scheme`, and/or `request.Host`, depending on what headers are passed. It doesn't appear that any code in the micromdm repository accesses any of those fields from the request, though I suppose some dependencies might (I didn't check - there are a lot!) This would be the easier approach, and could always be moved elsewhere if the need arises in the future.

I suppose you could copy the relevant code behind the handler to `groob/finalizer/logutil` to log extra fields, but I guess I prefer the composability of using it as middleware. If you do want the functionality in `logutil`, there wouldn't be a way put this functionality behind a flag (it's not recommended to parse proxy headers unless behind a trusted proxy) unless you change `logutil.NewHTTPLogger`'s signature or add `NewHTTPProxyLogger` or something simliar.

I don't really care either way. You and @groob just let me know what you'd like to see, and I can do a PR to groob/finalizer if need be.

jessepeterson (approved):

jessepeterson:
@korylprince looks like we have conflicts. tidy those up and we'll get this merged. i may open a tracking issue for future consideration of adding this to finalizer

korylprince:
@jessepeterson I got this rebased to micromdm:main. Thanks!

PR #745, opened 2021-04-26
Add Docker workflow. Add ARM build.

williamtheaker:
This adds a Github Actions workflow to build a multi-architecture Docker image when new release tags are added. Additionally, support for building ARM images is added.

Here's my [most recent action run](https://github.com/williamtheaker/micromdm/runs/2441081620?check_suite_focus=true) which generated builds for [Github's container registry](https://github.com/users/williamtheaker/packages/container/package/micromdm) and [Docker Hub](https://hub.docker.com/r/williamtheakergusto/micromdm/tags)

This assumes you have [enabled Github Container Registry for the micromdm organization](https://docs.github.com/en/packages/guides/enabling-improved-container-support#enabling-github-container-registry-for-your-organization-account) and added a Github Actions secret named `DOCKER_HUB_ACCESS_TOKEN` with a [Docker Hub token](https://docs.docker.com/docker-hub/access-tokens/#create-an-access-token).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,12 @@ FROM golang:latest as builder
 
 WORKDIR /go/src/github.com/micromdm/micromdm/
 
+ARG TARGETARCH
+ARG TARGETOS
+
 ENV CGO_ENABLED=0 \
-	GOARCH=amd64 \
-	GOOS=linux
+	GOARCH=$TARGETARCH \
+	GOOS=$TARGETOS
 
 COPY . .
 
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,40 @@
+name: Publish Docker image
+on:
+  release:
+    types: [published]
+jobs:
+  push_to_registry:
+    name: Build and push Docker image
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Set up QEMU for multi-architecture builds
+        uses: docker/setup-qemu-action@c308fdd69d26ed66f4506ebd74b180abe5362145 #v1.1.0
+        with:
+          image: tonistiigi/binfmt:latest
+          platforms: all
+
+      - name: Check out this repo
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
+        with:
+          fetch-depth: 1 # Checkout only latest commit
+
+      - name: Login to Github Packages Container registry with ephemeral token
+        run: docker login ghcr.io --username ${{ github.actor }} --password ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Login to Docker Hub with access token
+        run: echo ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }} | docker login --username micromdm --password-stdin
+
+      - name: Create builder instance
+        run: docker buildx create --use
+
+      - name: Build and push image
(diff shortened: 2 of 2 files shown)


jessepeterson:
I like the idea of encouraging diversification of container registries by not including Docker Hub (even at the added end-user expense/learning curve of not being able to "find" it easily). What do you think about removing the publish step to DH? In the end I would defer to @groob for the final verdict. :)

williamtheaker:
I'm all for decentralization. Heck I wish we were discussing this on Gitlab or [Sourcehut](https://sourcehut.org/). Centralizing image distribution by solely using GHCR doesn't seem like an effective measure to encourage the adoption of other image registries though.

From a purely technical aspect, Docker/Kubernetes tooling already privileges Docker Hub:
`If you don't specify a registry hostname, Kubernetes assumes that you mean the Docker public registry.` - [k8s documentation](https://kubernetes.io/docs/concepts/containers/images/).
If you run `docker info`, docker.io is the default registry. The same is true for Red Hat's `podman`.

By leaving up the Docker Hub images, we're adding an undocumented implicit distinction, where you need to know that a hostname prefix is required in your image tag in order to avoid downloading an image from February 2017. If you're not following Docker Inc's slow decay, it's difficult to tell which is correct `image: micromdm/micromdm:latest` vs `image: ghcr.io/micromdm/micromdm:latest`

Are there plans to [delete the micromdm Docker Hub repository](https://stackoverflow.com/a/63833570)? Otherwise this PR is adding a bit of automation to a footgun.

I see @khawaga and @nielshojen maintain micromdm images on quay.io and Docker Hub respectively, if they'd like to chime in.

jessepeterson:
> Centralizing image distribution by solely using GHCR doesn't seem like an effective measure to encourage the adoption of other image registries though.

How do you describe effective measures? At the very least omitting DH doesn't perpetuate the naive use of a de facto registry which I think is better than nothing.

> If you run docker info, docker.io is the default registry. The same is true for Red Hat's podman.

This is exactly my point. By not using those default registries it's a sort of forcing function for those less familiar with other registries to at least become familiar with alternatives.

> we're adding an undocumented implicit distinction

That's why we'd document it. :)

> Are there plans to delete the micromdm Docker Hub repository? Otherwise this PR is adding a bit of automation to a footgun.

Yes. We deleted the `micromdm/scep` DH repo already. I assumed MicroMDM was in line as well once alternatives were sorted. :)

williamtheaker:
> What do you think about removing the publish step to DH?
Done

Dockerfile
```diff
@@ -2,9 +2,12 @@ FROM golang:latest as builder
 
 WORKDIR /go/src/github.com/micromdm/micromdm/
 
+ARG TARGETARCH
```
jessepeterson, line 5: Do these have reasonable defaults or are they mandatory when doing a docker build?

williamtheaker: There are [implicit default values](https://docs.docker.com/engine/reference/builder/) for this already, which are overridden by the `--platform` flag in the buildx step.

Example Dockerfile:
```
FROM alpine
ARG TARGETPLATFORM
RUN echo "$TARGETPLATFORM"
```

Running this on an x86 machine running macOS:
```
docker build .
...
 => [2/2] RUN echo "linux/amd64"
 ```
 vs
 ```
docker build --platform linux/arm .
 ...
=> [2/2] RUN echo "linux/arm/v7"
``

PR #749, opened 2021-04-29
Add auto_advance_setup to DEP profile template

williamtheaker:
Added missing [DEP profile keys](https://developer.apple.com/documentation/devicemanagement/profile) to the template DEP profile produced by `mdmctl`.


jessepeterson (approved):
The template was never meant to be an exhaustive list (i.e. duplicating Apple's documentation). Just a 'getting started' point, really.  That said I see no  reason  not add these. Thanks!

PR #751, opened 2021-05-04
Ensuring MicroMDM Logos are Hidden to assistive technologies

bpmcneilly:
Following up on #721 (I was out when I got messages from @williamtheaker), I've added a few extras to the changes proposed there.

While the SVG a11y APIs should ensure that this SVG isn't exposed to assistive technologies, I've added `aria-hidden="true"` to ensure that the content is not exposed to users.

Similarly, while this bug is generally windows-based, and increasingly rare, some OS / browser combinations make all SVG elements focusable via the keyboard by default, so I've added `focusable="false"` to prevent that from happening in these odd corner cases.

Let me know if you have any questions here!

### DIFF
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -83,7 +83,7 @@ body {
 <body>
 
 <h1>MicroMDM</h1>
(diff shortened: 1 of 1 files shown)


jessepeterson (approved):
Thanks @bpmcneilly! Your insight here is super appreciated, even for our modest little landing page. :)

PR #754, opened 2021-05-27
Allow command callers to specify Command UUID for idempotency

daemonsy:
## What's up? 

Apple's MDM protocol allows a `CommandUUID` to be specified in commands for correlation during the handling of acknowledgement responses. 

Though it looks like an idempotency key, as far as I can tell, Apple doesn't do anything with it except passing it back to you in responses. 

Currently, MicroMDM generates a UUID when sending commands to Apple and returns it in the payload. 

### What this PR does 
This allows the the UUID to be specified by a API caller. e.g. in a `DeviceLock` call.

```json
{
    "request_type": "DeviceLock", 
    "pin": "123456", 
    "udid": "udid-of-the-device" ,
    "command_uuid": "This-is-my-UUID" 
}
```

When a `CommandRequest` is received, we accept the `command_uuid` set in the JSON payload or use the default generated one if it's not present. 

The intent for implementing it this way is to upstream the setting of the UUID as early as we can, so the rest of the code can always assume that the `CommandRequest.CommandUUID` is available, instead of making it optional. 

### Why do this? 
To ensure idempotency, my app has a command queue. Currently it sends a command to MicroMDM, takes the `command_uuid` and write a queue item that gets fulfilled by the ACK webhook event. But this is not safe because calling the MicroMDM API can fail or flake. 

It would make the command queue more robust to allow a stable command UUID to be written before the network calls to MicroMDM are made. For example, by allowing the caller to specify the key, even if a command succeeds but the response from MicroMDM is lost, the Apple webhook will still arrive with the right command UUID set. 

### Caveats 

After doing some testing, the main thing that nags me is that `""` is a valid value for a UUID. If you submit a JSON payload with `"command_uuid": ""`, we get back: 

```json
{
    "payload": {
        "command_uuid": "",
        "command": {
            "request_type": "DeviceLock",
            "pin": "123456"
        }
    }
}
```
This is because `encoding/json` treats `""` rightfully as a valid value and doesn't apply the `uuid.New.String()` default. 

The `DeviceLock` command works and I get back the ACK webhook, the `command_uuid` is **missing** and the plist payload has: 

```
<key>CommandUUID</key>
<string></string>
```

I'm happy to dive in a little deeper to use the default if the supplied value is an empty string or a whitespace only string

P.S. Also my first time writing Go code :smiley_cat:


jessepeterson:
Dup of #672 I think. :)

daemonsy:
Mmm looks pretty similar, #672 looks like it's abandoned though?

mdm/mdm/unmarshal_json.go
```diff
@@ -4,19 +4,22 @@ import (
...
 	"github.com/pkg/errors"
 )
 
 func (c *CommandRequest) UnmarshalJSON(data []byte) error {
 	var request = struct {
 		UDID        string `json:"udid"`
 		RequestType string `json:"request_type"`
-	}{}
+		CommandUUID string `json:"command_uuid"`
+	}{CommandUUID: uuid.New().String()}
```
jessepeterson, line 16: I think the right place for `uuid.New()` is in the `NewCommandPayload()` func. We can just test to see if its empty and fallback to a generated UUID if its empty. I saw your note on UUID working if blank, which of course is hilarious. We _could_ do that check with a string pointer for `CommandUUID`, above, but I'm not sure it's worth the added complexity (expounded a bit on this in #672).

mdm/mdm/command.go
```diff
@@ -17,7 +17,7 @@ type CommandPayload struct {
 
 func NewCommandPayload(request *CommandRequest) (*CommandPayload, error) {
 	payload := &CommandPayload{
-		CommandUUID: uuid.New().String(),
+		CommandUUID: request.CommandUUID,
```
jessepeterson, line 20: See below for notes on where `uuid.New()` should go and also whether this ought to be a pointer or not (I'd lean to no).

jessepeterson (changes requested):

jessepeterson:
In general I'm in favor of this. I added some notes. Welcome to Go! It's pretty standard for Go to require code formatting. This looks to be why the checks are failing — you ought to [`gofmt` your code (or better yet have your editor do it for you)](https://blog.golang.org/gofmt). :)

mdm/mdm/unmarshal_json.go
daemonsy: I think it's worth it to handle the complexity to check for empty strings. I cannot imagine any utility out of the it being empty, so setting it to the default MicroMDM generated UUID also prevents MicroMDM from having to deal with an  empty string internally that is supposed to be some tangible ID. 

I'll try for that in the simplest way possible.

jessepeterson: I think we're on the same page here but just to be clear:

> I think it's worth it to handle the complexity to check for empty strings. 

The complexity I refer to here is handling a pointer to a string in order to support usage of an empty UUID which I don't think is worth supporting. This is separate from simply checking for an empty supplied UUID (and defaulting to a generated UUID), which I'm in support of.

The thing here is to separate the concerns of JSON marshaling from default/supplied UUID handling (which probably belongs back in `NewCommandPayload()`).

daemonsy:
Hey @jessepeterson took a stab at your comments, also used gofmt. At least locally, seems like the command used by CI returns 0 line differences. 

I realized there's tests and wanted to add one, but it seems like the most relevant one is `mdm_command_test.go`. I sneaked a test into the `InstallProfile` case, which feels like a hijack. If you have advice on making a test case that is clearer, happy to do it.

mdm/mdm/mdm_command_test.go
```diff
@@ -204,7 +204,7 @@ func TestEndToEnd(t *testing.T) {
...
...VhZDg2ZDYxNTBkLmFsYWNhcnRlLmN1c3RvbXNldHRpbmdzLmZiOTU4ZDk2LWE0MzEtNDUzNi04NGQwLTFiZTQ4MjM4NWZiMjwvc3RyaW5nPgoJCQk8a2V5PlBheWxvYWRUeXBlPC9rZXk+CgkJCTxzdHJpbmc+Y29tLmFwcGxlLk1hbmFnZWRDbGllbnQucHJlZmVyZW5jZXM8L3N0cmluZz4KCQkJPGtleT5QYXlsb2FkVVVJRDwva2V5PgoJCQk8c3RyaW5nPmZiOTU4ZDk2LWE0MzEtNDUzNi04NGQwLTFiZTQ4MjM4NWZiMjwvc3RyaW5nPgoJCQk8a2V5PlBheWxvYWRWZXJzaW9uPC9rZXk+CgkJCTxpbnRlZ2VyPjE8L2ludGVnZXI+CgkJPC9kaWN0PgoJPC9hcnJheT4KCTxrZXk+UGF5bG9hZERlc2NyaXB0aW9uPC9rZXk+Cgk8c3RyaW5nPlN0b3BzIFNpcmkgZnJvbSBiZWluZyBlbmFibGVkLjwvc3RyaW5nPgoJPGtleT5QYXlsb2FkRGlzcGxheU5hbWU8L2tleT4KCTxzdHJpbmc+RGlzYWJsZSBTaXJpPC9zdHJpbmc+Cgk8a2V5PlBheWxvYWRJZGVudGlmaWVyPC9rZXk+Cgk8c3RyaW5nPkRpc2FibGVTaXJpPC9zdHJpbmc+Cgk8a2V5PlBheWxvYWRPcmdhbml6YXRpb248L2tleT4KCTxzdHJpbmc+PC9zdHJpbmc+Cgk8a2V5PlBheWxvYWRSZW1vdmFsRGlzYWxsb3dlZDwva2V5PgoJPHRydWUvPgoJPGtleT5QYXlsb2FkU2NvcGU8L2tleT4KCTxzdHJpbmc+U3lzdGVtPC9zdHJpbmc+Cgk8a2V5PlBheWxvYWRUeXBlPC9rZXk+Cgk8c3RyaW5nPkNvbmZpZ3VyYXRpb248L3N0cmluZz4KCTxrZXk+UGF5bG9hZFVVSUQ8L2tleT4KCTxzdHJpbmc+OWM3MzgwZDItNWJmZS00ZTYwLWJhZDMtMjVhZDg2ZDYxNTBkPC9zdHJpbmc+Cgk8a2V5PlBheWxvYWRWZXJzaW9uPC9rZXk+Cgk8aW50ZWdlcj4xPC9pbnRlZ2VyPgo8L2RpY3Q+CjwvcGxpc3Q+Cg==", "request_type": "InstallProfile"}`,
```
jessepeterson, line 207: I _love_ that you added a test,  that's great!  However, instead of replacing the existing test can you add another test (even  if its just a copy/paste)?

jessepeterson:
@daemonsy this looks good. If you want to add a different test and then add a line item describing this change in the `CHANGELOG` for unreleased then we can get it merged!

daemonsy:
Hey @jessepeterson Just got back to the computer world. I'm trying to add it as a separate test case. Will add the Changelog item too. Also, do you want me to rebase it into a single commit? Happy for you to squash merge the PR too.

jessepeterson (approved):
Would also like an end-to-end test (like you had) in addition to the unit tests but this looks good I think. Feel free to submit another PR for that test if you're feeling up to it. :)

daemonsy:
https://github.com/micromdm/micromdm/pull/756

Issue #735, opened 2021-03-12
Bootstrap Token


meta-github:
### What version of `micromdm` are you using?
1.8.0

### What `micromdm` command did you run?
On the DEP enrolled client: `profiles status -type bootstraptoken`

### What did you expect to see?
micromdm should support the bootstrap token protocol.

### What did you see instead?
`profiles: Bootstrap Token supported on server: NO`

Bootstrap token (introduced by macOS Catalina) help to get secure token to user accounts and to authorize software updates (which makes a big difference on Apple Silicon). If I currently issue a ScheduleOSUpdate command to a DEP enrolled client running macOS 11.2.3 on Apple Silicon, that commands will not get executed unless it is locally authorized by an admin account with secure token. Botstap Token support would be a great help to get **unattended software updates** on the Apple Silicon platform.

From the MDM point of view, Bootstrap Token is a set of commands the MDM server must be able to handle. It is not some kind of MDM payload or configuration. The protocol is described by Apple as part of the [Check-in](https://developer.apple.com/documentation/devicemanagement/check-in) command.

jessepeterson:
Just a note that you can trivially fool the OS into thinking the MDM server supports bootstrap tokens by adding `com.apple.bootstraptoken` to the `ServerCapabilities` array in the MDM enrollment profile payload. But, as you noted there are specific `MessageTypes` to handle as part of the MDM check-in protocol.

_Technically_ MicroMDM may not need to "support" the tokens so much as make sure they are passed on to the Webhook (for any integrations to handle) but we may want to integrate them into the device storage for "native" support, too.

tomaswallentinus:
The second sounds like a great idea. If micromdm doesn't recognise the endpoint or payload it could just forward it to the webhook.

ivanhata:
I'd really much like to see this implemented way or another. So +1 defenitely.

I've been playing with the idea that I'd give this a try. Lack of go skills are holding me back a bit at least. Also my understanding of micromdm's internal design could be better :)

Anyway, what would be the best places to put the different parts of the implementation? Some would be quite obvious and copy-paste-replace could be enough, maybe. But especially the responses the server should be sending to the client. To my understanding, this is a bit different from all the previous where the server is usually the one sending requests and the client responding. Now it's the opposite. I haven't been able to find a suitable example from the existing code.

ivanhata:
I started working with this and got that far that the client escrows the bootstrap token and it gets saved to micromdm's database.

[...]

PR #757, opened 2021-06-09
Added SetRecoveryLock support

tomaswallentinus:
It says SetFirmwarePassword on the files, but its meant to be SetRecoveryLock and VerifyRecoveryLock.

I was thinking about test files, but in what file should I put the tests?

PR #760, opened 2021-06-14
Bootstrap Token

ivanhata:
_Rebased to current main and corrected the branch, otherwise this is identical with closed #755_ 

This is a partial implementation of Bootstrap Token as discussed in #735
I had a short chat with @jessepeterson. He recommended doing a PR to get this worked on further.

Fetching the token from the db and sending it back to the client when the client asks for it ie sends the GetBootstrapTokenRequest is not implemented. Everything else should work.

Also earlier code for Bootstrap Token from commit 58c3782 "mdm: add SetBootstrapToken command." is removed

I have spent some time with the code trying to understand and figure out a way to implement what is missing. I'm quote stucked now. If someone wants to continue from this and implement what is needed, please do so, I'll be just happy to get the feature 🙂. Anyhow, I'm ready to continue, but I'd need help. Any tips or advice is appreciated.


mdm/internal/checkinproto/checkin.pb.go
```diff
@@ -1,11 +1,15 @@
-// Code generated by protoc-gen-go. DO NOT EDIT.
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
```
jessepeterson, line 1: This doesn't seem right. Seems like `protoc-gen-gogo` generated this file but this file should be generated with normal `protoc-gen-go`. How did you create this file? `go generate` should be all that's needed once the required `protoc` and `protoc-gen-go` is installed.

mdm/mdm/command.go
```diff
@@ -66,7 +66,6 @@ type Command struct {
 	ScheduleOSUpdateScan            *ScheduleOSUpdateScan
 	ActiveNSExtensions              *ActiveNSExtensions
 	RotateFileVaultKey              *RotateFileVaultKey
-	SetBootstrapToken               *SetBootstrapToken
```
jessepeterson, line 69: Thanks for removing this, too!

mdm/service.go
```diff
@@ -14,10 +14,12 @@ type Service interface {
...
-	AuthenticateTopic = "mdm.Authenticate"
-	TokenUpdateTopic  = "mdm.TokenUpdate"
-	CheckoutTopic     = "mdm.CheckOut"
+	ConnectTopic           = "mdm.Connect"
+	AuthenticateTopic      = "mdm.Authenticate"
+	TokenUpdateTopic       = "mdm.TokenUpdate"
+	CheckoutTopic          = "mdm.CheckOut"
+	GetBootstrapTokenTopic = "mdm.GetBootstrapToken"
+	SetBootstrapTokenTopic = "mdm.SetBootstrapToken"
 )
```
jessepeterson, line 24: Below here, line 31 in the `MDMService` is where we'll couple our device service to the check-in service (using a simplified API, hopefully).

platform/device/internal/deviceproto/device.pb.go
```diff
@@ -1,21 +1,15 @@
-// Code generated by protoc-gen-go.
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
```
jessepeterson, line 1: Here, too, I think this needs to be generated with `protoc-gen-go` instead of `-gogo`. I'm surprised you got this if you just ran `go generate`.

platform/device/worker.go
```diff
@@ -179,6 +191,45 @@ func (w *Worker) updateFromCheckout(ctx context.Context, message []byte) error {
 
 }
 
+func (w *Worker) updateFromGetBootstrapToken(ctx context.Context, message []byte) error {
```
jessepeterson, line 194: So I'm not sure there's any need to listen for a _get_ request. Because we need to actively return data there's nothing really to "listen" for here. Indeed it only looks like we update the LastSeen and/or Await state here. I think this can go in lieu of coupling the device service to the check-in service, mentioned elsewhere.

jessepeterson (changes requested):
This all looks really good!

The primary thing I think we need to do is to couple a new "BootstrapTokenGetter" (or similar) interface to the `MDMService` which knows how to retrieve the Bootstrap token. We can't really do this over the pub/sub interface because the request-response model here doesn't really fit the pub/sub paradigm. Basically we'll make the device service implement this new "BootstrapTokenGetter" interface and when we init the MDM interface we'll provide it this service (which is actually just the device interface).

I hope that makes sense and let me know of any questions!

jessepeterson:
I put together #764 which will probably help for the bootstrap token retrieval.

jessepeterson:
hey @ivanhata i've merged the other PR. this sets you up to do the work mentioned in the review after a rebase. no rush though, just letting you know. :)

ivanhata:
Sorry for the horrible delivery time!

These commits should bring the earlier partial implementation to fully working feature. I have done only some testing, like
- watching both micromdm and mdmclients logs in debug mode
- toggling the state of bootstrap token with profiles command
- evaluating the output of sysadminctl, diskutil and fdesetup for bootstrap and secure token

Let me know if further testing is needed in general or something needs to be tested more thoroughly.

Huge thank you to @jessepeterson for his patience and advice.

mdm/internal/checkinproto/checkin.pb.go
ivanhata: I had some trouble to run just _go generate_ . I don't recall what I did exactly.

I tried to do everything by the book in the later commit. Still couldn't get the go generate to exit without errors. After some googling found a hint to add extra option to the proto file. After doing that, go generate exited ok.

I noticed there was some discussion in #757 about same kind of difficulties.

platform/device/worker.go
ivanhata: I forgot to do anything for this one. I need to check the code again if I figure out how to change it. And if you think that this is still a good idea.

platform/device/builtin/db.go
```diff
@@ -46,6 +47,24 @@ func NewDB(db *bolt.DB) (*DB, error) {
...
 
+func (db *DB) GetBootstrapToken(ctx context.Context, udid string) ([]byte, error) {
+	opts := device.ListDevicesOption{}
+	
+	var udids []string
+	udids = append(udids, udid)
+
+	opts.FilterUDID = udids
+	
+	devices, err := db.List(ctx, opts)
```
jessepeterson, line 58: We should avoid `db.List` here. It enumerates every device and only filters after the fact. We should instead use `db.DeviceByUDID()` which then obviates the need for all the UDID stuff.

platform/device/builtin/db.go
```diff
@@ -46,6 +47,24 @@ func NewDB(db *bolt.DB) (*DB, error) {
...
+	for _, d := range devices {
+		btstring, _ := hex.DecodeString(string(d.BootstrapToken))
```
jessepeterson, line 60: Any reason to store/encode/decode as hex? We can store the raw binary in the device record.

platform/device/builtin/db.go
```diff
@@ -46,6 +47,24 @@ func NewDB(db *bolt.DB) (*DB, error) {
 	return datastore, nil
 }
 
+func (db *DB) GetBootstrapToken(ctx context.Context, udid string) ([]byte, error) {
```
jessepeterson, line 50: nit: maybe return b64Data from the other definition?

```suggestion
func (db *DB) GetBootstrapToken(ctx context.Context, udid string) (B64Data, error) {
```

jessepeterson (changes requested):
this is great progress! in addition to the code changes I think you'll need to: rebase the PR on main as well as confirm working & testing (though it sounds like @korylprince might be willing to help test).

I suspect the biggest pain is going to be not changing the protobuf files/definitions (we want to consolidate that into #773). this implies running some older protoc and protobuf compiler versions.

ivanhata:
I guess we can just close this PR as everything is solved in #781 now.

PR #764, opened 2021-06-18
Refactor MDM service to return a payload from check-ins

jessepeterson:
Currently all of the Checkin messages don't return a response—they all succeed without any return value. However, certain Check-in messages require returning a payload such as `UserAuthenticate`, `GetBootstrapToken`, `DeclarativeManagement`, etc.

This PR adjusts the check-in interface to allow for returning payload data on a Check-in message which gets returned to the client verbatim (much like the `Acknolwedge` interface method already does). Note this PR does not implement any of the above; rather it just makes it possible for them to be implemented.

### DIFF
--- a/mdm/checkin.go
+++ b/mdm/checkin.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Checkin is an MDM Service check-in
 func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) ([]byte, error) {
 	// reject user settings at the loginwindow.
 	// https://github.com/micromdm/micromdm/pull/379
--- a/mdm/checkin.go
+++ b/mdm/checkin.go
@@ -10,31 +10,31 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) error {
+func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) ([]byte, error) {
 	// reject user settings at the loginwindow.
 	// https://github.com/micromdm/micromdm/pull/379
 	if event.Command.MessageType == "UserAuthenticate" {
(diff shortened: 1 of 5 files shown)


discentem:
Hey Jesse, 
Can you reword your description of what you are changing/implementing? I unfortunately don't understand what you mean by 

> This will allow the various check-in messages which return a payload to operate

mdm/checkin.go
```diff
@@ -10,31 +10,31 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) error {
+func (svc *MDMService) Checkin(ctx context.Context, event CheckinEvent) ([]byte, error) {
```
discentem, line 13: Thoughts on adding a function comment to explain what the returned `[]byte` represents and can be used for?

discentem: I realize there might be many functions which don't currently adhere to this standard of commenting but I think it would still be of value here. 

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_always_document_public_symbols

jessepeterson: Sure. I met you half-way. All of the implementors of the interface I add a small comment to. But I also added documentation to the main interface.

jessepeterson:
> Hey Jesse,
> Can you reword your description of what you are changing/implementing? I unfortunately don't understand what you mean by
> 
> > This will allow the various check-in messages which return a payload to operate

Done, hopefully more clear now.

mdm/checkin.go
jessepeterson: I walked this back, actually. In the document you posted:

> There is one exception to this rule; you don’t need to document methods that implement an interface. Specifically don’t do this:

So I removed those. :)

discentem (approved):
LGTM

discentem:
Note, I haven't yet tested this commit myself.

PR #773, opened 2021-08-15
Remove gogo/protobuf

jessepeterson:
Attempt to cleanup protobuf usage per the notes in #742. But the big one is that gogo/protobuf is unmaintained/looking for project ownership so revert to Google's protobuf library.

* Remove gogo/protobuf from all sources
* Make all `protoc` compiler runs consistent and modern:
  * Change the few `go generate` files using `--gofast_out=` to the normal compiler
  * Change all `go generate` to use `--go_opt=paths=source_relative` flag to keep our invocation/directory layout
  * Change all `.proto` files to include `option go_package = "...";` (compiler fails otherwise)
* Regenerate all protobuf files using modern `protoc` and `protoc-gen-go`
* Run tests `make test`
* Finally, `go mod tidy`

This needs some reasonable amount of testing to make sure everything works as expected.


jessepeterson:
Was able to perform a command queue on the main branch (w/o this PR), then de-queue using this PR seemingly fine. Also @korylprince put some testing into this PR, too.

PR #778, opened 2021-09-29
update pkcs7 dependency

jessepeterson:


jessepeterson:
Includes a potential fix for the ber2der error. see micromdm/nanomdm#32.

PR #781, opened 2021-10-05
Bootstrap Token

korylprince:
I hope I'm not stepping on @ivanhata's toes (:heart:). I've taken his existing work in #760 and made the following changes to hopefully get this PR merged:

* Rebased on main
* Reverted unnecessary changes to proto files and generated files
  * Note: @ivanhata changed three proto files, and each file needed a different protoc compiler version to match existing code :man_facepalming:. Check the commit for a note on the versions used
* Fixed some spacing/syntax/naming issues pointed out by @jessepeterson and some I noticed as well
* Cleaned up the writing/reading of the Bootstrap token (no unnecessary conversions to hex or base64)
* Added a small test for `GetBootstrapToken`
* Fixed the GetBootstrapToken checkin event not being send to the webhook
  * Note: the `[]byte` returned from `MDMService.Checkin` isn't sent to pubsub/webhook. Not sure if that's something that would be wanted, but that's probably for another PR. I'm not sure it could be done without breaking compatibility somewhere.

I've done some testing with a few different MacBooks:

* Ran `sudo profiles install -type bootstraptoken` then `sudo profiles validate -type bootstraptoken` to verify proper escrow
  * Note: `sudo profiles status -type bootstraptoken` will lie to you and tell you the server has the token just because it sent it. You need to use `validate` to verify the MacBook can actually get the token from the server and use it properly.
 * Verified messages are passing properly via webhook and HTTP debug logs
 * Verified users get a SecureToken when they login after escrowing the Bootstrap Token
   * a `GetBootstrapToken` request is sent each time a user without a SecureToken logs in
   * this works even if you're using NoMAD Login AD :tada:

As a final note, this PR addresses all of @jessepeterson's reviews on #740 except for his note on `platform/device/worker.go: Worker.updateFromGetBootstrapToken`. From my understanding, you're saying to move the logic from the worker to `mdm/checkin.go: MDMService.Checkin`. I guess where I'm confused, is why you wouldn't do the same thing for `updateFromCheckout`, `updateFromTokenUpdate`, etc. It seems like pretty much everything in the work is just accessing the db and pubsub.

In any case, I can make any changes to this PR to get it merged, so just let me know!


jessepeterson (approved):
This looks good. Much thanks to you and @ivanhata!

> each file needed a different protoc compiler version to match existing code

Yikes, how annoying. Thanks for slogging through! Mentioned in #760 I really want to unify all that with #773; but it'll take time & testing.

> Added a small test for GetBootstrapToken

❤️ 

> Note: the []byte returned from MDMService.Checkin isn't sent to pubsub/webhook. Not sure if that's something that would be wanted, but that's probably for another PR. I'm not sure it could be done without breaking compatibility somewhere.

Yeah the issue here is, conceptually, the webhook publishes the Checkin request messages of which GetBootstrapToken is just that — a request. We don't really publish Checkin responses — which actually were only added recently. This is true of any Checkin, or Acknowledge (command) requests, too. We publish the request, but not the MDM server response. There could be another PubSub topic, I suppose, for GetBootstrapToken *responses*, but as you said: probably another discussion from another PR (if it's even something that's worth doing). It [works the same way in NanoMDM](https://github.com/micromdm/nanomdm/blob/cc53d2812187825f4f83945deff4fbc03b59e878/service/microwebhook/service.go#L103) fwiw, too.

> except for his note on platform/device/worker.go: Worker.updateFromGetBootstrapToken. From my understanding, you're saying to move the logic from the worker to mdm/checkin.go: MDMService.Checkin. I guess where I'm confused, is why you wouldn't do the same thing for updateFromCheckout, updateFromTokenUpdate, etc. It seems like pretty much everything in the work is just accessing the db and pubsub.

Thanks, yeah. This was a misconception from the original suggestion to include the 'full' device database in the Checkin service when, in fact, we only have a BootstrapTokenRetriever interface. If it were the full device database then we wouldn't need to publish to the PubSub because we'd have the device right there (which we don't, of course). So, that's my bad. And as you note the code in the PR is actually consistent with the other checkin requests, too. So: ignore me, carry on. :)

----

Again, thanks for this! I'll have another look-see hopefully tomorrow and we can get this in!

ivanhata:
This is great! Thanks to @korylprince for fixing all the issues and getting this done! I'm totally fine with it if this PR gets merged. I guess the #760 can be just closed.

mdm/checkin_event.go
```diff
@@ -60,6 +63,17 @@ type userTokenUpdate struct {
 	NotOnConsole  bool   `plist:",omitempty"`
 }
 
+// Get Bootstrap Token Message Type # GetBootstrapTokenRequest
```
jessepeterson, line 66: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

mdm/checkin_event.go
```diff
@@ -60,6 +63,17 @@ type userTokenUpdate struct {
...
+type getBootstrap struct {
+	GetAwaitingConfiguration bool
+}
+
+// Set Bootstrap Token Message Type # SetBootstrapTokenRequest
```
jessepeterson, line 71: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

mdm/service.go
```diff
@@ -21,25 +21,33 @@ type Service interface {
...
-	CheckoutTopic     = "mdm.CheckOut"
+	ConnectTopic           = "mdm.Connect"
+	AuthenticateTopic      = "mdm.Authenticate"
+	TokenUpdateTopic       = "mdm.TokenUpdate"
+	CheckoutTopic          = "mdm.CheckOut"
+	GetBootstrapTokenTopic = "mdm.GetBootstrapToken"
+	SetBootstrapTokenTopic = "mdm.SetBootstrapToken"
 )
 
+type BootstrapTokenRetriever interface {
```
jessepeterson, line 32: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

platform/device/builtin/db.go
```diff
@@ -46,6 +46,14 @@ func NewDB(db *bolt.DB) (*DB, error) {
 	return datastore, nil
 }
 
+func (db *DB) GetBootstrapToken(ctx context.Context, udid string) ([]byte, error) {
```
jessepeterson, line 49: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

platform/device/device.go
```diff
@@ -38,6 +38,7 @@ type Device struct {
 	DEPProfileAssignedDate time.Time        `db:"dep_profile_assigned_date"`
 	DEPProfileAssignedBy   string           `db:"dep_profile_assigned_by"`
 	LastSeen               time.Time        `db:"last_seen"`
+	BootstrapToken         string           `db:"bootstrap_token"`
```
jessepeterson, line 41: Any reason this, too, shouldn't be a byte slice? I see a few `string()` and `[]byte()` type conversions and I wonder if they're necessary.

jessepeterson (approved):
Just a couple really minor things, but overall looking good!

mdm/checkin_event.go
korylprince: I updated these to match existing code. For example `auth struct` has the comment `// Authenticate Message Type`, not `auth Message Type`. These are private structs so they're not going to show up in godoc anyways. If you want it to actually be `//getBootstrap ...` I can do that too.

mdm/checkin_event.go
korylprince: same as above

platform/device/device.go
korylprince: I mainly left these as is because it matches existing code (namely, Token and Unlock Token are also stored as strings). I agree it would make more sense to store it as a byte (same in the protobuf), but figured consistency with the rest of the codebase was better. []byte-string conversions cost very little I would think. If you want to move it to a []byte, I can make those changes (proto file, generated code, services, etc). Just let me know.

jessepeterson: Sure, sounds good. :)

PR #782, opened 2021-10-05
Use []byte instead of string for BootstrapToken protobuf

korylprince:
This fixes issues that occur when rebasing #773 onto current main. Storing non-UTF-8 characters in a protobuf string doesn't cause issues for the version of protoc-gen-go currently used for device.proto in main, but it does cause [this issue](https://github.com/golang/protobuf/issues/1228) with the updated version used for #773. 

This PR just changes the BootstrapToken to be stored as a []byte, and removes any string/[]byte conversions like @jessepeterson and I had discussed on #781.

I've tested this and verified functionality is the same as #781  (`profiles install, validate`).

I've also tested that rebasing #773 on top of this PR doesn't have [this issue](https://github.com/golang/protobuf/issues/1228). (That's the only testing I've done on #773 so far.)

Also, as an aside, Device.Token and Device.UnlockToken are also stored as strings in the protobuf, but they're encoded to hex first. Maybe this issue is why...


jessepeterson:
It looks like the underlying type is still a string? I vote that we do one of two things: we convert the protobuf type to a `bytes` type (over a string) or we encode (hex/b64/whatever) the value like we do with `Device.Token` and `Device.UnlockToken` (as you noted). Thoughts?

korylprince:
I can make that change if you want, just let me know. In practice, protos generated when it was a string still read correctly with the new code. Just let me know and I can add a commit to the PR.

jessepeterson (approved):
lgtm!

PR #785, opened 2021-11-12
implement mdmctl apply app -sign-identity

korylprince:
This PR adds support for signing packages on other OSes (any OS that supports the `xar` command) using a .p12 identity file.

The following flags are added to `mdmctl apply app`:

* `-sign-identity` - path to .p12 developer identity
* `-password` - password for identity (can also be empty)
* `-out` - output path for signed package. This isn't necessarily related to multi-OS support, but is very useful and solves package overwriting if the original pkg is in `/tmp`. 

Help strings have been updated for the new flags.

I did not end up adding tests as they would have mainly just replicated the tests in go-macos-pkg. I did manually test basically every permutation of the new flags to verify functionality.

Note: if you're having weird issues trying to build, make sure you're on up to date Go. I hit [this bug](https://github.com/golang/go/issues/47979) when pulling in  go-macos-pkg.


korylprince:
Finally got the builders happy enough to build... That was a much longer journey than I anticipated.

groob (approved):

PR #791, opened 2021-12-20
Show mdmctl config switch usage if no server -name passed

williamtheaker:
`mdmctl config switch` currently doesn't provide any hint that a `-name` flag should be passed:

```
% ./mdmctl config switch
no server named "" found
% ./mdmctl config switch example
no server named "" found
```

After this patch:
```
% build/darwin/mdmctl config switch
USAGE
  mdmctl config switch [flags]

FLAGS
  -name   name of the server to switch to
```

### DIFF
--- a/cmd/mdmctl/config.go
+++ b/cmd/mdmctl/config.go
@@ -274,6 +274,9 @@ func switchServerConfig(name string) error {
 	if err != nil {
 		return err
 	}
+	if name == "" {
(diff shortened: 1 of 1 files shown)


jessepeterson (approved):

jessepeterson (changes requested):
Fix format and gtg! :)

jessepeterson (approved):
lgtm!

PR #792, opened 2022-01-10
Fix to 502 Bad Gateway error while communicating with Apple Servers

HernanPaez:
Hi! Today I was renewing my certificates and I found an error while signing the push certificate:

```
$ ./mdmctl mdmcert vendor -sign -cert=./mdm-certificates/mdm.cer -password=MyAwesomePassword

signing push certificate request with vendor private key: load root certificate from http://www.apple.com/appleca/AppleIncRootCertificate.cer: got 502 Bad Gateway when trying to http.Get http://www.apple.com/appleca/AppleIncRootCertificate.cer
```

I tried to get the certfile using wget using the same url and it throws the same error. The only way to retrieve that file was to communicate using https.

I know this pull request is trivial (I only added an 's' to the url) but I was able to create and sign my push certificate after this change.

### DIFF
--- a/pkg/crypto/mdmcertutil/certutil.go
+++ b/pkg/crypto/mdmcertutil/certutil.go
@@ -73,7 +73,7 @@ func (p *PushCertificateRequest) Encode() ([]byte, error) {
 
 const (
 	wwdrIntermediaryURL = "https://www.apple.com/certificateauthority/AppleWWDRCAG3.cer"
-	appleRootCAURL      = "http://www.apple.com/appleca/AppleIncRootCertificate.cer"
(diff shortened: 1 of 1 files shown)


korylprince:
For what it's worth, it seems like this was probably a temporary error, as the http:// link works for me now. However, it's probably a good idea to get the certificate over HTTPS regardless.

HernanPaez:
Well, that's weird. It was down all day yesterday. I suppose that they were waiting for the PR :) I should have waited one more day to renew the certs. 

Feel free to close this PR if you want. 👍

jessepeterson (approved):
Agreed. While maybe not technically required we should default HTTPS. This is also what is linked from https://www.apple.com/certificateauthority/

PR #795, opened 2022-01-14
Add goreleaser to improve build process

networkpanic:
This PR introduces [goreleaser](https://goreleaser.com/) as a build tool for Golang releases.

Other changes:
- universal2 binaries for Darwin targets
- Dockerfile built in a minimal approach
- Changelog generation for Releases

Goreleaser creates new releases as soon as a git tag was pushed.


korylprince:
You might want to disable the linter (or make it warn only if possible) as it's failing on quite a few errors.

networkpanic:
It is allowed to fail to unblock builds while keeping the errors to make us aware

gavinelder:
The main concern I have is that the body of this PR is quite large and covers a number of different concerns

For example 

The addition of the `golangci-lint` step as a GH action could generate enough discussion to be a separate PR.

The moving of the released Docker image from Alpine -> Scratch could have environmental impacts which would need additional testing and validation, potentially releasing multiple flavours of an image distribution would be a less impactful change.

jessepeterson:
Thanks @gavinelder for those comments. I concur. :)

PR #802, opened 2022-02-20
Added RefreshCellularPlans support

ayush5harma:
Add ability to send RefreshCellularPlans mdm-command:

**About the Refresh Data Plans command**
The Refresh Data Plans command is sent from the MDM solution to the device and provides the address of the network provider’s eSIM (SM-DP+) server. The device then downloads, installs and activates its eSIM. It may take up to three minutes for the installation and activation to occur.

https://developer.apple.com/documentation/devicemanagement/refreshcellularplanscommand


go.sum
```diff
@@ -20,6 +20,8 @@ github.com/go-sql-driver/mysql v1.4.0/go.mod h1:zAC/RDZ24gD3HViQzih4MyKcchzm+sOG
 github.com/go-stack/stack v1.6.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
 github.com/go-stack/stack v1.7.0 h1:S04+lLfST9FvL8dl4R31wVUC/paZp/WQZbLmUgWboGw=
 github.com/go-stack/stack v1.7.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
+github.com/gogo/protobuf v1.3.2 h1:Ov1cvc58UF3b5XjBnZv7+opcTcQFZebYjWzi34vdm4Q=
```
jessepeterson, line 23: Looks like somehow we've added back gogo/protobuf? We'll want to keep that removed.

jessepeterson (changes requested):
I think there might be some additional code that's missing here — the marshal and unmarshal_proto.go files for instance.

jessepeterson (changes requested):
looks like build actions aren't passing. `go mod tidy` seems like.

PR #803, opened 2022-02-23
Add support for Priority in ScheduleOSUpdate.

meta-github:
This pull request adds support for the property `Priority` that was recently added to macOS 12.3 (beta). Documentation is https://developer.apple.com/documentation/devicemanagement/scheduleosupdatecommand/command/updatesitem.

Setting a `High` priority (default is `Low`) hopefully results in faster starting downloads.


jessepeterson (approved):
ltgm, thanks!

PR #809, opened 2022-03-18
Added RefreshCellularPlans support

ayush5harma:
Add ability to send RefreshCellularPlans mdm-command:

**About the Refresh Data Plans command**
The Refresh Data Plans command is sent from the MDM solution to the device and provides the address of the network provider’s eSIM (SM-DP+) server. The device then downloads, installs and activates its eSIM. It may take up to three minutes for the installation and activation to occur.

https://developer.apple.com/documentation/devicemanagement/refreshcellularplanscommand


jessepeterson (approved):
Overall lgtm, haven't done an in-depth review nor am I able to test this so: thanks for including a unit test!

PR #819, opened 2022-05-10
clarify mdmcert.download instructions

discentem:


korylprince:
The mdmcert.download addition LGTM. I think we'll need to rework the top section under `Configure an APNS certificate` to make it more clear that `mdmcert.download` and `Generate MDM CSR` are mutually exclusive.

jessepeterson:
So.. there's something to be said here for _not_ mentioning mdmcert.download so as to sorta.. nudge folks to looking at obtaining an official certificate. I figure that's the original reason why they were excluded from these docs.

jessepeterson:
Also - I think the docs are missing a critical step in the 'MDM CSR' workflow, too. In the CLI docs there's three steps but in the MD docs here its missing one (mdm push cert generation, I think).

jessepeterson:
Finally: I'm not blocking this issue on that, just bringing up those points ;P

discentem:
> The mdmcert.download addition LGTM. I think we'll need to rework the top section under `Configure an APNS certificate` to make it more clear that `mdmcert.download` and `Generate MDM CSR` are mutually exclusive.

I moved this to a collapsable section in order to make it more clear that they are mutually exclusive. 

> So.. there's something to be said here for _not_ mentioning mdmcert.download so as to sorta.. nudge folks to looking at obtaining an official certificate. I figure that's the original reason why they were excluded from these docs.

I was under the impression you were okay with adding more docs around mdmcert.download. If you'd rather not then that's fine too.

discentem:
> Also - I think the docs are missing a critical step in the 'MDM CSR' workflow, too. In the CLI docs there's three steps but in the MD docs here its missing one (mdm push cert generation, I think).

Is this relevant to the mdmcert.download bits? If not, can someone address this in a separate PR?

korylprince:
I like the collapsed section. The one issue I see left is that the new mdmcert.download section stops with downloading the certificate, when really the last step should be uploading it. You could add that as step 8, or you could break out the last section of `Generate MDM CSR` into it's own heading (level ###) and point to it.

Per @jessepeterson , the `Generate MDM CSR` is missing this step from the CLI docs:

```
Next, create a push CSR. This step generates a CSR required to get a signed a push certificate.

	mdmctl mdmcert push -password=secret -country=US -email=admin@acme.co
```

It would go before the `Sign a push certificate...` section. @discentem you can include it in this PR if you want, or I can add it after this is merged.

Assuming @jessepeterson has no qualms about it, I'll merge this once the above issue is addressed.

discentem:
> I like the collapsed section. The one issue I see left is that the new mdmcert.download section stops with downloading the certificate, when really the last step should be uploading it. You could add that as step 8, or you could break out the last section of `Generate MDM CSR` into it's own heading (level ###) and point to it.
> 
> Per @jessepeterson , the `Generate MDM CSR` is missing this step from the CLI docs:
> 
> ```
> Next, create a push CSR. This step generates a CSR required to get a signed a push certificate.
> 
> 	mdmctl mdmcert push -password=secret -country=US -email=admin@acme.co
> ```
> 
> It would go before the `Sign a push certificate...` section. @discentem you can include it in this PR if you want, or I can add it after this is merged.
> 
> Assuming @jessepeterson has no qualms about it, I'll merge this once the above issue is addressed.

I've split out upload into its own section as requested. Please re-review.

korylprince (approved):
LGTM

PR #824, opened 2022-05-25
Add max_user_deferrals

tomaswallentinus:
The maximum number of times the system allows the user to postpone an update before it’s installed. The system prompts the user once a day.
This key is only supported when InstallAction is InstallLater and only supported for minor OS updates (for example, macOS 12.x to 12.y).

PR #828, opened 2022-06-04
Updated schedule_os_update to support more parameters.

meta-github:
Resolve #824 and add other missing options in ScheduleOSUpdate command.

I have tried to make this script compatible to the existing one that uses three positional arguments.


jessepeterson (approved):
Neat! I haven't tested it, but looks like it should work. Let's roll with it. :)

PR #833, opened 2022-06-27
Change PushCertificateRequest file extension to txt

moogle19:
Since the Apple Push Certificates Portal lately only allows `.csr`, `.txt` or `.rtf` files, the `.plist` file cannot be uploaded to the Portal (see Screenshot). 

By changing it to .txt it works like expected.

Screenshot 2022-06-27 at 12 34 33

### DIFF
--- a/cmd/mdmctl/mdmcert.go
+++ b/cmd/mdmctl/mdmcert.go
@@ -53,7 +53,7 @@ Once you created the push CSR, you mush sign the push CSR with the MDM Vendor Ce
 
     mdmctl mdmcert vendor -sign -cert=./mdm-certificates/mdm.cer -password=secret
 
-Once generated, upload the PushCertificateRequest.plist file to https://identity.apple.com to obtain your MDM Push Certificate.
+Once generated, upload the PushCertificateRequest.txt file to https://identity.apple.com to obtain your MDM Push Certificate.
 Use the push private key and the push cert you got from identity.apple.com in your MDM server.
 
 Commands:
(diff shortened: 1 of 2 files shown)


moogle19:
Okay. If it is a problem on Apples side, I will open a ticket there.

But maybe it would be nice to add a little comment to the documentation, like "if you cannot upload the plist file, change the extension to .txt" or something like that.

jessepeterson:
This is definitely Apple's bug (again). For a little humor: https://twitter.com/wikiwalk/status/1504099614873509892 :)

jessepeterson:
I think Apple has regressed on this changed and may not be willing to fix it. Can someone confirm that the APNs portal still requires the file to be renamed? If so then I think we should probably get this merged.

petarov:
Seems so. I just tried to upload a CSR and got the error:

`You have uploaded an invalid file type. Supported file extensions are .txt, .rtf, .plist, .b64`

It's a client side validation at https://identity.apple.com/pushcert/js/createNewExtCertificateView.js, Line: 288  in `#isValidFileExtension()`.  I've noticed that they actually extended the method. Previously it was just:

```
function isValidFileExtensionAndSize() {
    var MAX_FILE_SIZE = 1 * 1024 * 1024; // 1MB
    var ALLOWED_FILE_EXTENSIONS = /(\.txt|\.rtf)$/I;
```

To me this is a sign that it won't get "fixed".

jessepeterson:
I see you updated your comment but when I visit that URL I see:

`var ALLOWED_FILE_EXTENSIONS = /(\.txt|\.rtf|\.plist|\.b64|\.csr)$/i;`

Which should make it good to go for us. I'll close this again. :)

petarov:
So, Apple seem to have modified the validation again last Thursday. Now there's no file extension check at all. 🤷‍♂️ 

 ```js
 function isValidFileExtension() {
    return true;

	/* var ALLOWED_FILE_EXTENSIONS = /(\.txt|\.rtf|\.plist|\.b64|\.csr)$/i;

	var fileInput = document.getElementById("upfile");
	var filePath = fileInput.value;
    if(!ALLOWED_FILE_EXTENSIONS.exec(filePath)){
	         return false;
    }
	return true; */
}
```

PR #850, opened 2022-12-19
Make dep client thread safe

petitout:
When the client instance is shared between. multiple go routines, creating a session can result in 429 from Apple

### DIFF
--- a/dep/client.go
+++ b/dep/client.go
@@ -53,7 +53,7 @@ type Client struct {
 	userAgent string
 	client    HTTPClient
 
-	baseURL *url.URL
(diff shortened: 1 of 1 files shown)


korylprince (changes requested):
This LGTM. Can you run gofmt (`gofmt -w .` in the root of the project) to appease the CI testers? Pushing a new commit or rebasing the change and force-pushing it will update this PR.

korylprince (approved):

korylprince:
Merged after discussion with @jessepeterson.

PR #855, opened 2023-01-24
Manually verify SHA1 certificates (fixes #854)

### LINKED ISSUE #854: unauthorized enrollment client
1.10.0
No command, DEP client tries to enroll.
Enrollment should succeed.
### What did you see instead?
On the server:
```
transport=http err="unauthorized enrollment client: not signed by Apple Device CA"
transport=http method=POST status=500 proto=HTTP/1.1 host=16.181.2.145 user_agent=CloudConfiguration-OSX/1.0 path=/mdm/enroll
```
On the client:
```
2023-01-23 09:47:14.548520+0100 0x3e16     Default     0x0                  336    0    mdmclient: (ConfigurationProfiles) [com.apple.ManagedClient:ManagedClient] CloudConfig: Get activation profile returning error after 0.25 secs
2023-01-23 09:47:14.548640+0100 0x3e16     Default     0x0                  336    0    mdmclient: (ConfigurationProfiles) [com.apple.ManagedClient:ManagedClient] CloudConfig: Error returning activation profile.   Error = 500; Datasize = 0
```
Then, no DEP enrollment.

I have no idea what the server is checking and how I can turn it off or make it successful.

Client and server are running macOS 13.1.

Switching back to my older micromdm binary (1.9.0, compiled about a year ago; eventually patched regarding ScheduleOSUpdate) has fixed the issue.

korylprince:
Per discussions on the macadmins slack, this changes crypto.VerifyFromAppleDeviceCA to use a manual signature verification instead of using Go's x509 library, since SHA1WithRSA is deprecated and planned to be removed from Go.

I also went ahead and stuck the public key in a var so it doesn't have to parse the embedded cert on each call.


pkg/crypto/apple.go
```diff
@@ -91,19 +93,23 @@ pyQPbp8orlXe+tA8JA==
 -----END CERTIFICATE-----
 `
 
+// extracted public key from appleiPhoneDeviceCAPEM
+var appleiPhoneDeviceCAPublicKey = &rsa.PublicKey{
```
jessepeterson, line 97: I'd like to keep this within the single X.509 PEM that's already in the source. If optimization was the concern we could add an `init()` function that does the parsing just once at startup time (though I generally dislike them). I know you test for equality in the test but I like the single source of truth better. Thoughts? Could also wrap this into an interface with its own initializer as well.

korylprince: I thought the test was sufficient, but I can see how it seems a little "dirty".

I pushed a commit that replaces the handcrafted public key with a "mustParsePublicKey" function. This way, we're still only parsing it once, but we have the single source of truth. Does that work for you?

jessepeterson: Yeah this looks good!

I don't think its "dirty" per se — I just dislike multiple sources of truth. Like if we wanted to update that cert or whatever we only have one place we need to do it. (Though, honestly, if Apple did update that cert we'd probably have to go do something else yet again heh).

PR #857, opened 2023-02-02
Enable Dependabot version updating.

williamtheaker:
Someone (me) pinned old versions of Github Actions, which is causing deprecation warnings in your actions: `The save-state command is deprecated and will be disabled soon. Please upgrade to using Environment Files.` & `Node.js 12 actions are deprecated.`. This PR enables Dependabot, which will checking for updates and open PRs to bump the pinned version for Github Actions and Go modules.

I've left the checkout action in the Docker worklow one version behind, so you can see an example Dependabot PR upon merging this branch.

I also moved `gofmt` to earlier in the CI run, on the assumption that you don't want to waste minutes building commits that haven't been linted.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Don't change this despite the path being .github/workflows
+    schedule:
+      # Check for updates to GitHub Actions on the first day of the month
+      interval: "monthly"
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      # Check for updates to Go modules on the first day of the month
+      interval: "monthly"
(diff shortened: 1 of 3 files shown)


jessepeterson:
This looks good and I'll take a closer look in a bit. But I'm curious: where do you see the deprecation warnings? When I look at the last run actions I don't (immediately) see any problems (all green checks). Anyway thanks!

williamtheaker:
Scroll to the bottom of a run like this, to the Annotations section:
https://github.com/micromdm/micromdm/actions/runs/3998500583

```
 Build, test & format (1.16.x, ubuntu-latest)
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-go@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

jessepeterson (approved):
lgtm!

PR #864, opened 2023-02-18
Add raw command and clear device queue endpoints

korylprince:
Following discussions on the recent office hours, this PR adds support for enqueuing raw plist commands. This PR also adds an endpoint to clear command queues, since a malformed command could break a command queue. Without a way to clear a device's queue, a command queue could become unfixable, short of manually editing the database.

Here's a summary of the changes:

**Raw Command Enqueuing**

* Add POST /v1/command/{udid} endpoint
  * The request body is just the raw command plist
  * This is similar to NanoMDM's command enqueuing endpoint, though it doesn't currently support multiple ids
* Internally, this is published to a new `command.RawCommandTopic`, which the boltdb and inmem queues have new pollers for
 
**Clear Device Queue**

* Add DELETE /v1/command/{udid} endpoint
* This uses the existing `queue.Clear` method used during enrollment
* Internally, this couples the Command service with the command queue, as opposed to using the pubsub, but there is precedent with the command queue being coupled to the MDM service
  * I think this is acceptable since this endpoint should be used sparingly

**Note:** the code is complete, but I've marked as WIP since I haven't tested it yet.

### DIFF
--- a/platform/command/clear.go
+++ b/platform/command/clear.go
@@ -11,7 +11,7 @@ import (
 )
 
 func (svc *CommandService) ClearQueue(ctx context.Context, udid string) error {
-	if err := svc.queue.Clear(context.TODO(), mdm.CheckinEvent{Command: mdm.CheckinCommand{UDID: udid}}); err != nil {
+	if err := svc.queue.Clear(ctx, mdm.CheckinEvent{Command: mdm.CheckinCommand{UDID: udid}}); err != nil {
 		return errors.Wrap(err, "clearing command queue")
 	}
 	return nil
--- a/platform/command/clear.go
+++ b/platform/command/clear.go
@@ -1,0 +1,47 @@
+package command
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/gorilla/mux"
+	"github.com/micromdm/micromdm/mdm"
+	"github.com/pkg/errors"
+)
+
+func (svc *CommandService) ClearQueue(ctx context.Context, udid string) error {
+	if err := svc.queue.Clear(context.TODO(), mdm.CheckinEvent{Command: mdm.CheckinCommand{UDID: udid}}); err != nil {
+		return errors.Wrap(err, "clearing command queue")
+	}
+	return nil
+}
+
+type clearRequest struct {
+	UDID string
+}
+
+type clearResponse struct {
+	Err error `json:"error,omitempty"`
+}
+
+func (r clearResponse) Failed() error   { return r.Err }
+func (r clearResponse) StatusCode() int { return http.StatusOK }
+
+func decodeClearRequest(ctx context.Context, r *http.Request) (interface{}, error) {
+	return clearRequest{UDID: mux.Vars(r)["udid"]}, nil
+}
+
+// MakeClearQueueEndpoint creates an endpoint which clears device queues.
+func MakeClearQueueEndpoint(svc Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(clearRequest)
+		if req.UDID == "" {
+			return clearResponse{Err: errEmptyRequest}, nil
+		}
+		if err := svc.ClearQueue(ctx, req.UDID); err != nil {
+			return clearResponse{Err: err}, nil
+		}
+		return clearResponse{}, nil
+	}
+}
--- a/platform/command/new_command.go
+++ b/platform/command/new_command.go
@@ -2,6 +2,8 @@ package command
 
 import (
 	"bytes"
+	"fmt"
+	"io"
 	"net/http"
 
 	"github.com/go-kit/kit/endpoint"
@@ -118,19 +120,19 @@ func decodeNewRawCommandRequest(ctx context.Context, r *http.Request) (interface
 		return nil, errors.New("empty udid")
 	}
 
-	buf := new(bytes.Buffer)
-	if _, err := buf.ReadFrom(r.Body); err != nil {
-		return nil, err
+	payload, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, fmt.Errorf("read payload body: %w", err)
 	}
 
 	// verify body is valid plist and parse CommandUUID and RequestType
 	var req newRawCommandRequest
-	if err := plist.NewXMLDecoder(buf).Decode(&req); err != nil {
-		return nil, err
+	if err := plist.NewXMLDecoder(bytes.NewBuffer(payload)).Decode(&req); err != nil {
+		return nil, fmt.Errorf("parse payload as plist: %w", err)
 	}
 
 	req.UDID = udid
-	req.Raw = buf.Bytes()
+	req.Raw = payload
 	return req, nil
 }
 
--- a/platform/command/new_command.go
+++ b/platform/command/new_command.go
@@ -1,9 +1,12 @@
 package command
 
 import (
+	"bytes"
 	"net/http"
 
 	"github.com/go-kit/kit/endpoint"
+	"github.com/gorilla/mux"
+	"github.com/groob/plist"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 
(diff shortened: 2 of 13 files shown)


korylprince:
After a few bugfixes, and a decent amount of testing, this seems to work consistently and correctly.

**Testing:**

* Start with MDM joined device that's offline
* [Inspect command queue](https://gist.github.com/korylprince/4952a2949170b10c8fd0ffcbdd289889) to see it's empty
* Queue raw command with raw_command tool
* Inspect command queue to see command was added
  * Repeat to verify multiple commands can be queued
* Clear the queue with clear_queue tool
* Inspect command queue to see commands (and NotNows) have been cleared
* Queue some more command(s) with raw_command tool
* Bring device online
* Confirm device responding to commands (e.g. via webhook)
* Inspect command queue to verify queue is empty (and Completed queue has grown)
* Attempt to queue some malformed commands (missing RequestType, malformed plist) and verify they are rejected

platform/command/new_command.go
```diff
@@ -69,3 +92,65 @@ func MakeNewCommandEndpoint(svc Service) endpoint.Endpoint {
...
+func (r newRawCommandResponse) Failed() error   { return r.Err }
+func (r newRawCommandResponse) StatusCode() int { return http.StatusCreated }
+
+func decodeNewRawCommandRequest(ctx context.Context, r *http.Request) (interface{}, error) {
+	udid, ok := mux.Vars(r)["udid"]
+	if !ok {
+		return nil, errors.New("empty udid")
+	}
+
+	payload, err := io.ReadAll(r.Body)
```
jessepeterson, line 123: close body?

platform/command/new_command.go
```diff
@@ -69,3 +92,65 @@ func MakeNewCommandEndpoint(svc Service) endpoint.Endpoint {
...
+	if err != nil {
+		return nil, fmt.Errorf("read payload body: %w", err)
+	}
+
+	// verify body is valid plist and parse CommandUUID and RequestType
+	var req newRawCommandRequest
+	if err := plist.NewXMLDecoder(bytes.NewBuffer(payload)).Decode(&req); err != nil {
```
jessepeterson, line 130: any reason not to just directly decode the body rather than buffer it? I guess to log it? it'll still generate an error if we parse it directly tho?

```suggestion
	if err := plist.NewXMLDecoder(r.Body).Decode(&req); err != nil {
```

jessepeterson (commented):
Also our comments on the call. :)

platform/command/new_command.go
```diff
@@ -69,3 +92,65 @@ func MakeNewCommandEndpoint(svc Service) endpoint.Endpoint {
...
+var errMalformedRequest = errors.New("request is malformed")
+
+// MakeNewRawCommandEndpoint creates an endpoint which creates new raw MDM Commands.
+func MakeNewRawCommandEndpoint(svc Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(newRawCommandRequest)
+		if req.UDID == "" {
+			return newRawCommandResponse{Err: errEmptyRequest}, nil
+		}
+		if req.CommandUUID == "" || req.Command.RequestType == "" {
```
jessepeterson, line 148: maybe also check for an empty/nil payload?

platform/queue/queue.go
```diff
@@ -318,6 +322,58 @@ func (db *Store) pollCommands(pubsub pubsub.PublishSubscriber) error {
 	return nil
 }
 
+func (db *Store) pollRawCommands(pubsub pubsub.PublishSubscriber) error {
```
jessepeterson, line 325: any reason not to re-use the existing command poller (and event, pubsub topic, etc)?

jessepeterson (commented):
Thanks for looking into this! I'm in favor of this feature. Got a couple questions (which I'm sure you have answers to; probably just me not looking at this code recently enough!).

platform/command/new_command.go
korylprince: While I can't find anywhere that says this explictly, I believe MDM doesn't support binary plist commands? So this is explicitly only parsing XML plists (and also skipping the binary plist check that `Unmarshal` has).

platform/command/new_command.go
```diff
@@ -69,3 +92,65 @@ func MakeNewCommandEndpoint(svc Service) endpoint.Endpoint {
...
+		return nil, errors.New("empty udid")
+	}
+
+	payload, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, fmt.Errorf("read payload body: %w", err)
+	}
+
+	// verify body is valid plist and parse CommandUUID and RequestType
+	var req newRawCommandRequest
```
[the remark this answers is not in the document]
korylprince: Not really. I was just following [the convention in the existing code.](https://github.com/korylprince/micromdm/blob/rawcmd/platform/command/new_command.go#L74) Also, `req` is not returned as a pointer (again, following convention in the existing code), so we'd have to deference it on return if it was declared as a pointer.

platform/command/clear.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http"
+
+	"github.com/go-kit/kit/endpoint"
+	"github.com/gorilla/mux"
+	"github.com/micromdm/micromdm/mdm"
+	"github.com/pkg/errors"
+)
+
+func (svc *CommandService) ClearQueue(ctx context.Context, udid string) error {
+	if err := svc.queue.Clear(context.TODO(), mdm.CheckinEvent{Command: mdm.CheckinCommand{UDID: udid}}); err != nil {
```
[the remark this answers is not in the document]
korylprince: Good catch. Not sure where why I did that...

platform/command/new_command.go
jessepeterson: Ah, I see. As an extra-implicit check? Fair 'nuff.

Complete aside: Profiles (at least the initial OTA profile) can be bplists. A popular vendor uses that. fwiw.

platform/command/new_command.go
korylprince: The request (type `newRawCommandRequest`) is parsed by `decodeNewRawCommandRequest` before being handed to the Endpoint. This parses the body as a plist, so it guarantees that the payload is non-nil and contains raw plist, though not necessarily a properly formatted command.

platform/queue/queue.go
korylprince: I considered sharing the poller, but ultimately thought keeping them separate made it a bit cleaner. If you would rather have them combined in one select (and use one less goroutine), I can make that change easily.

Using a separate event type and pubsub topic allows things to stay *much* simpler. To use the existing `Event` type, we'd have to add a `RawPayload []byte` field, since it doesn't contain a `[]byte` field. Then we'd have to add more logic to the marshal/unmarshal code to differentiate between the two, probably by adding another field (e.g. `type`) to the underlying event protobuf.

By having separate event types, we can keep the same underlying protobuf untouched (since it *does* use a plain []byte field), and differentiate between the two by using two separate topics.

Another thing to consider is that the existing event/queue is real-world tested for several years. While I think the raw command code is solid, if there is some bug, keeping it segregated in another type and topic makes sure it doesn't need to change the existing, working system.

korylprince:
I finally got around to adding the [api docs](https://github.com/korylprince/micromdm/blob/rawcmd/docs/user-guide/api-and-webhooks.md#schedule-raw-commands-with-the-api). Assuming those look good, and you're good with my responses to your latest code review, this should be ready to merge!

platform/command/new_command.go
jessepeterson: The decoder could also guarantee the command uuid and request tyke are also set tho, yeah? Negating the need for those checks?

platform/queue/queue.go
jessepeterson: Yep you're 100% correct here. I completely forgot we not only marshal/unmarshal command JSON, but in fact marshal it into protobuf. And the current pubsub event is dedicated to that marshaling. The sepratet event and poller make perfect sense in that context. I just forgot about it, lol. :)

jessepeterson (approved):

platform/command/new_command.go
korylprince: For sure. Again, it mostly goes back to matching existing code (e.g. the non-raw command endpoints) which:

PR #865, opened 2023-02-19
adding a script to install VPP apps

krmzbbr:
Installing VPP apps require a different format than install_appstore_application. itunes_store_id (as number ) is used instead of identifier. purchase_method option is also required.


krmzbbr:
> Can you add a comment to this command that stipulates that you need to have VPP/Apps&Books configured and setup in order to use this command?

Comment added.

jessepeterson (approved):

PR #871, opened 2023-03-25
Address issues with /mdm/checkin endpoint

korylprince:
This PR is to address some issues discovered while [troubleshooting enrollment issues](https://macadmins.slack.com/archives/C19RTE0L9/p1679701827459239) on the MacAdmins Slack:

* The device cert pkcs7 verification currently used on the /mdm/checkin doesn't account for time skew
  * Since the device is generating the certificate on the fly, if the device is just a few seconds ahead of the server, the server current time will be outside of the certificate validity, and verification will fail
  * This is fixed by adding a 5 minute skew with [VerifyWithChainAtTime](https://pkg.go.dev/go.mozilla.org/pkcs7#PKCS7.VerifyWithChainAtTime)
* Most of the errors in the mdm Endpoints are silently dropped because the response encoder called the error encoder directly, instead of just returning the error so it would pass back through the error logger.
  * This is fixed by making the response encoder returning errors.

### DIFF
--- a/mdm/server.go
+++ b/mdm/server.go
@@ -138,8 +138,7 @@ func encodeResponse(ctx context.Context, w http.ResponseWriter, response interfa
 	}
 
 	if e, ok := response.(failer); ok && e.Failed() != nil {
-		encodeError(ctx, e.Failed(), w)
-		return nil
+		return e.Failed()
 	}
 
 	w.WriteHeader(http.StatusOK)


mdm/server.go
```diff
@@ -138,8 +139,7 @@ func encodeResponse(ctx context.Context, w http.ResponseWriter, response interfa
 	}
 
 	if e, ok := response.(failer); ok && e.Failed() != nil {
-		encodeError(ctx, e.Failed(), w)
```
jessepeterson, line 141: I think by not calling `encodeError()` we're breaking certain MDM functionality by not returning appropriate HTTP status information from specific errors. Maybe a better option is still call `encodeError()` but have it return the error if it doesn't match the others, else nil — in this way we can both pass up the error (if need be) or handle our special MDM status codes?

jessepeterson (commented):
So my observation here is that in however many years we've been going we've never had the timestamp issue come up on cert verification. Further I'd want to determine where 5 minutes comes from — did we just pick that? I think the SCEP server tries to issue certs from -10 minutes ago — but of course that's for new enrollments. And then if we did want an arbitrary time in there we should make it a documented `const`.

I'd probably learn towards _not_ wanting to add a specific verification timestamp exemption — especially if errors will be properly logged now (i.e. the case that brought this up would probably have been caught).

mdm/server.go
jessepeterson: Oh, but perhaps this is what `httptransport.ServerErrorEncoder(encodeError),` does? It's already called for us? In that case, yeah, I see the issue.

jessepeterson: Looking at go-kit now, we also could have chosen to use StatusCoder-abiding errors to handle this logic, it seems.

korylprince: The logic in go-kit [calls encodeError](https://github.com/go-kit/kit/blob/7f14cb4dc16cd4005dc7770c131ae685cc4943b7/transport/http/server.go#L132-L134) if the response encoder returns an error.

So currently (before this PR), the response encoder calls encodeError directly and returns nil, bypassing the error logging in go-kit. Now we just return the error in the response encoder, which go-kit then logs and calls encodeError directly after.

korylprince:
Yeah the 5 minutes was a bit arbitrary, but a fairly safe bet. For instance, this is what [Microsoft recommends](https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/maximum-tolerance-for-computer-clock-synchronization). I added a commit to move it to a documented var.

I'm also surprised this is the first time we've seen it. I'm not sure, but it seems like the cert in question wasn't the SCEP cert, but a cert generated by the device (so maybe this was OTA enrollment? I don't really have any experience with that). The biggest reason I think that is because how else would this situation arise? If it was a SCEP cert, it would be generated by the server, so it wouldn't really matter what the device's time was. I think this would also answer the question of why it hasn't been seen yet: most people using MicroMDM are using DEP, so they _would_ just use the SCEP cert to begin with. This particular issue occurred when someone was enrolling an Apple TV with Configurator.

Apologies if I'm not understanding these processes correctly; I've read through the wiki and the OTA code, but I'm still not 100% clear on the process. Would it even make sense to see a device-generated cert on the /mdm/checkin endpoint?

Assuming I am correct above, then adding a skew (of 5 minutes or whatever) makes a lot of sense. If a device is generating it's own cert on the fly, if it's just a few seconds ahead of the server, then it will be out of the validity period. When @tgunz synced his server with NTP, it was only 2 seconds offset, and that was enough to allow the device to enroll.

mdm/server.go
korylprince: Resolved per our discussion during office hours.

korylprince:
To recap our discussion of this PR during office hours today:

- The issue is _not_ to do with devices generating their own certificates (in both the normal and OTA enrollments, the device uses a SCEP cert)
- The issue _is_ because the device is generating pkcs7 signatures for requests, and clock skew between the device and the server is causing the signatures to be invalid
- Because the issue seems to occur rarely, the clock skew is now behind an adjustable (default off) flag

jessepeterson (approved):

PR #882, opened 2023-05-01
Declarative Management hand-off

jessepeterson:
NanoMDM has been able to "hand-off" the Declarative Management protocol to another server for [some time now](https://github.com/micromdm/nanomdm/releases/tag/v0.3.0) using the [`-dm` flag](https://github.com/micromdm/nanomdm/blob/main/docs/operations-guide.md#-dm). This PR brings the same capability to MicroMDM with a similar `-dm` command line flag. This feature brought was brought up in #759 for MicroMDM.

This allows you to use a Declarative Management "server" like [KMFDDM](https://github.com/jessepeterson/kmfddm)* with devices enrolled in MicroMDM.See also [this blog post](https://micromdm.io/blog/wwdc21-declarative-management/) about initial DDM protocol testing.

This work banks off the work that @korylprince has done in #864 to target raw commands by allowing us to submit raw DeclarativeManagement commands to MicroMDM.

\* *Note:* KMFDDM needs the newest commit to target MicroMDM's command API (by using the `-micromdm` switch with it). Please also read the KMFDDM Quickstart guide/tutorial for getting started.


discentem:
> * Note: KMFDDM needs the newest commit to target MicroMDM's command API (by using the -micromdm switch with it). Please also read the KMFDDM Quickstart guide/tutorial for getting started.

Can you link the commit? 😅

jessepeterson:
@discentem sure, it's 

> > * Note: KMFDDM needs the newest commit to target MicroMDM's command API (by using the -micromdm switch with it). Please also read the KMFDDM Quickstart guide/tutorial for getting started.
> 
> Can you link the commit? 😅

Sure: jessepeterson/kmfddm@186f5dd7f0d610862023d67f2393103f0ed2c5a6 :)

jessepeterson:
With reviews of this PR we can try and roll releases of both MicroMDM and KMFDDM with these features in the next week or two.

server/dm.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	req.Header.Set(enrollmentIDHeader, id)
+	if len(data) > 0 {
+		req.Header.Set("Content-Type", "application/json")
+	}
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
```
korylprince, line 65: ioutil.ReadAll is deprecated in favor of io.ReadAll.

Note: there are 9 other instances of ioutil.ReadAll in the codebase 🤷

server/dm.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
```
korylprince, line 69: Should this come before reading from resp.Body? e.g. reading from the Body could fail, which would skip the Close. I don't see any guarantees from net/http that such a case wouldn't still need Close called.

platform/device/udidauth.go
```diff
@@ -94,23 +94,19 @@ func (mw *udidCertAuthMiddleware) Checkin(ctx context.Context, req mdm.CheckinEv
...
-		return nil, errors.Errorf("unknown checkin message type %s", req.Command.MessageType)
 	}
+	matched, err := mw.validateUDIDCertAuth([]byte(req.Command.UDID), hashCertRaw(devcert.Raw))
+	if err != nil {
+		return nil, err
+	}
+	if !matched && !mw.warnOnly {
+		return nil, errors.New("device certifcate UDID mismatch")
+	}
+	return mw.next.Checkin(ctx, req)
```
korylprince, line 111: You're removing a check for invalid message types here. Just confirming this is because this check also happens in the actual checkin handler (e.g. mw.next)?

korylprince (approved):
A couple of notes, but nothing show-stopping. LGTM!

server/dm.go
jessepeterson: Nice catch! I moved it up 4 lines. Also a copy/paste from NanoMDM. Not sure how that made it in. I'll change there, too.

platform/device/udidauth.go
jessepeterson: Yep. For the purposes of UDID<->cert auth we don't really care what the messages are — just that initial authenticate message is handled separately. This removes a duplicate check that has to be maintained whenever messages changed (like they did for BS tokens, and now DM).

jessepeterson:
Addressed your HTTP client comments @korylprince. Also applied those fixes in micromdm/nanomdm@3abd0dc82a3cb40e2fc4da42a2d9659570da9507 :shrug:

Issue #886, opened 2023-05-31
What is "error retrieving device certificate: CMS verify MDM Signed Message" and how to fix it?


GregOriol:
### What version of `micromdm` are you using?
v1.11.0

### What `micromdm` command did you run?
micromdm serve
then trying to enroll a device

### What did you expect to see?
The same device used for my tests did enroll before many times, but after an erase command and a new set up today, enrolling fails with this error

### What did you see instead?
In micromdm's logs, I get exactly this:
```
err="error retrieving device certificate: CMS verify MDM Signed Message: pkcs7: signing time \"2023-05-31T11:51:17Z\" is outside of certificate validity \"2023-05-31T11:51:19Z\" to \"2024-05-30T11:51:19Z\""
transport=http method=PUT status=500 proto=HTTP/1.0 host=127.0.0.1 user_agent=MDM/1.0 path="/mdm/checkin"
```

It looks like there are a few seconds of difference somewhere causing the issue, but I don't understand where and how to fix this.

The iPad is freshly reset up and is set to automatic time.
The server is time synced with ntp.

Could you help?

hslatman:
Hey @GregOriol, a similar issue was recently discussed on the `micromdm` channel in the MacAdmins Slack workspace. One user reported that ensuring the device and the server had its time set correctly to resolve the issue. You mention you've done that, so I suspect a difference between the NTP servers in use. Not sure if it's possible to make it such that both point to the same NTP server?

A PR was also opened: https://github.com/micromdm/micromdm/pull/871 that discussed this, but not all of its changes were merged. Here is a branch that still has the option for a configurable clock skew: https://github.com/korylprince/micromdm/tree/pkcs7-skew.

hslatman:
Great, @GregOriol 😄 

I guess it makes sense to have this option for a (slight) skew available.

korylprince:
I've opened a [new PR](https://github.com/micromdm/micromdm/pull/887) to add this flag. I was never able to replicate this issue, but it's obvious it's happening to multiple people. For now, you can build and test against https://github.com/korylprince/micromdm/tree/pkcs7-skew. I'm going to reopen this issue until that PR is closed.

PR #887, opened 2023-05-31
Add configurable clock skew flag

korylprince:
This PR is just cherry-picked from the work I previously did in #871.

At the time, I was unable to replicate the clock skew issue, and we only had one report. Since then, there's been a couple of issues ([NanoMDM #71](https://github.com/micromdm/nanomdm/issues/73), #886) reported experiencing the same issue @tgunz first reported on the MacAdmins Slack.

As-is, this adds a flag to configure the clock skew, and it defaults to 0s.

I still think it's worth exploring setting the default to 5 minutes (see [here](https://github.com/micromdm/micromdm/pull/871#issuecomment-1483955814) for why that duration). It would mean the issue is fixed transparently for most users instead of issues being raised here or on the MacAdmins slack.

More importantly, I don't see any downsides to this change. In theory, I guess this could possibly open you to easier replay attacks, but the attacker would have to MiTM your client to get the header in the first place, and that would be a much larger issue. Maybe I'm missing something else here, though...

### DIFF
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/micromdm/micromdm/mdm"
 	"github.com/micromdm/micromdm/mdm/enroll"
+	"github.com/micromdm/micromdm/pkg/crypto"
 	httputil2 "github.com/micromdm/micromdm/pkg/httputil"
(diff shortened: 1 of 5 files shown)


korylprince:
Rebased this back onto main to catch up with other PRs.

jessepeterson (approved):
lgtm

PR #890, opened 2023-06-05
add flag to include timestamp in logs

korylprince:
There currently doesn't appear to be a way to have MicroMDM include a timestamp in it's logs. This PR adds a flag (default off) that will set the "ts" log key to the UTC timestamp.


jessepeterson (approved):

jessepeterson:
from slack:

> I always thought this was purposeful because the thing you’re sending the logs to would do its own timestamping (and, indeed, when run out of, say, systemd it does that for you). but i can’t think of a good reason not to have a switch to turn it on :shrug:. so thanks again kory!

PR #902, opened 2023-07-11
Chore: tidy for go 1.20

jamesez:
Go 1.20's `test` will fail mdmctl because of strings with trailing newlines. This PR removes the newlines, and adds 1.20 to the build matrix.

I also tidied three minor complaints from staticcheck.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -11,7 +11,7 @@ jobs:
     name: Build, test, and format
     strategy:
       matrix:
-        go-version: [1.16.x, 1.17.x, 1.20.x]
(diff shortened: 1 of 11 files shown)


jamesez:
> i notice a few (most?) of the newline changes are removing newlines from text that is directly output. i.e. it’s for formatting/output style on a console. does your PR maintain that same formatting?

I've made all but one out be md5 identical to previous - the previous output had a trailing whitespace on one line.

korylprince (approved):
LGTM. I also updated the Go versions for CI and the Dockerfile, and updated the CHANGELOG.

PR #911, opened 2023-08-02
Adding capability to pass VPNUUID while sending InstallApplication command

paragpatel:


jessepeterson:
Hello @paragpatel I'd be fine with this addition if you added a test case. Thanks!

jessepeterson (changes requested):
If you could add a test case per above comment that'd be great!

PR #967, opened 2024-02-04
Allow disabling :80 to :443 redirect (Fixes #963)

### LINKED ISSUE #963: binding to port 80 cannot be altered or turned off when using builtin tls
micromdm version v1.11.0

Just started the server without turning off the tls functionality, i.e. **not** using `-tls=false`.

The server should not bind to port 80.

### What did you see instead?
The server binds to port 80 and that is neither optional nor configurable. There should be an option to turn off that http-bind or it should be eliminated completely.

korylprince:
This PR adds a flag to allow disabling the :80 to :443 redirect. This PR is pending merge of https://github.com/micromdm/go4/pull/16 and updating go.mod/go.sum.


jessepeterson (approved):
lgtm!

meta-github:
Thank you for this change. I have switched our server to a version compiled from main and it no longer binds to port 80!

PR #975, opened 2024-03-03
bugfix: Adjust error comparison in builtin stores

iann0036:


jessepeterson (approved):
I'd love to see a regression test, but given all this Bolt code is leaning toward legacy, it's fine. Nice catch!

PR #1054, opened 2025-01-22
Fix #641: Unable to disable bluetooth

### LINKED ISSUE #641: Boolean field false is omitted
When I pass this body in the url v1/commands

```
    body: JSON.stringify({
      udid: tokens[0],
      request_type: "Settings",
      settings: [{
        "item": "Bluetooth",
        "enabled": false
      }]
    })
```

In the plist result the field Enabled is omitted, it works when the value is true.

I think the problem is  in the library github.com/groob/plist at this line:
https://github.com/groob/plist/blob/a99fbe489d03c90cfeff2db2643d13916e9a3898/encode.go#L129

Thank you in advance

ashutosh624:
Current behaviour:
The Settings payload works for enabling bluetooth on the device, but it fails to disable it.
```
{ "request_type": "Settings", "udid": "<uid>", "Settings": [
    { "Item": "Bluetooth", "Enabled": true }
]}
```

As mentioned in the issue #641, the nil values are omitted while serialising the payload into plist.
This PR removes conversion of `false` to nil value.

### DIFF
--- a/mdm/mdm/unmarshal_proto.go
+++ b/mdm/mdm/unmarshal_proto.go
@@ -414,7 +414,8 @@ func protoToSetting(s *mdmproto.Setting) Setting {
 		setting.Enabled = nilIfFalse(pbs.GetEnabled())
 	case "Bluetooth":
 		pbs := s.GetBluetooth()
-		setting.Enabled = nilIfFalse(pbs.GetEnabled())
+		setting.Enabled = new(bool)
+		*setting.Enabled = pbs.GetEnabled()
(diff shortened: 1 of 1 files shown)


ashutosh624:
@jessepeterson Please take a look.

jessepeterson:
@ashutosh624 could you please create a test case for this (like the other command test cases), would love to see it mocked up. I'm concerned this always sets it false even if it's missing from the input.

ashutosh624:
@jessepeterson Added tests. Please take a look.

PR #1078, opened 2025-05-05
fix(queue): in memory concurrent r/w panic

drkr-sl:
Added mutex to in memory queue for `q.queue` manipulations as this was causing a lot of panics


jessepeterson (approved):
lgtm, nice catch. Thanks!

PR #1096, opened 2025-06-20
add auto setup admin accounts changes

harryvik990:


natewalck:
What is the purpose of this diff? Don't throw up a diff with a bunch of changes with zero context 😁

PR #1124, opened 2026-07-03
[MDMv2]  Add opt-in /mdm/migration-checkin endpoint for NanoMDM migration dual-write

simar811:
## Summary

Adds a Basic-Auth-protected `PUT /mdm/migration-checkin` endpoint, gated behind a new
`-migration` flag (off by default), that lets an external dual-writer inject check-in
events into MicroMDM's device/enrollment store **without** a device certificate.

This exists to support the live-fleet **MicroMDM → NanoMDM migration**, where devices are
moved to NanoMDM via per-device DNS. Once moved, a device's fresh `TokenUpdate` (new APNs
push token from a reboot/re-enroll/token rotation) only reaches NanoMDM — MicroMDM's copy
goes stale. If we later roll a device back to MicroMDM, MicroMDM would push against a stale
token and the device is silently unreachable until it happens to check in on its own.

With this endpoint, `mdmdirector` mirrors every `Authenticate`/`TokenUpdate`/`CheckOut`
webhook it receives from NanoMDM into MicroMDM, keeping MicroMDM fresh enough to remain a
safe rollback target for the whole migration window.

## How it works

The real `/mdm/checkin` path can't be reused: it requires a valid `Mdm-Signature` header
(only the device can produce it) and is wrapped in cert-auth middleware that rejects — and
in one case panics — when no device certificate is present.

The new handler - parses the raw check-in plist into a CheckinCommand, and pushes it to the pubsub bus.

This mirrors NanoMDM's own `-migration` endpoint (Basic Auth, skips cert verification), which
`llorne` already uses to seed NanoMDM in the forward direction.

### DIFF
--- a/cmd/micromdm/serve.go
+++ b/cmd/micromdm/serve.go
@@ -130,6 +130,7 @@ func serve(args []string) error {
 		flLogTime                = flagset.Bool("log-time", false, "Include timestamp in log messages")
 		flP7Skew                 = flagset.Int("device-signature-skew", env.Int("MICROMDM_DEVICE_SIGNATURE_SKEW", 0), "Sets the allowable clock skew (in seconds) when verifying device signatures")
 		flDisableRedirect        = flagset.Bool("disable-redirect", env.Bool("MICROMDM_DISABLE_REDIRECT", false), "disable the :80 -> :443 redirect if listening on :443")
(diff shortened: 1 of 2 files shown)


simar811:
Closing — targeting my own fork instead.

PR #1125, opened 2026-07-09
Add opt-in -migration flag support for micromdm migration endpoint

simar811:
## What

Wires up the new micromdm `-migration` server flag (added in micromdm
`Add migration sync endpoint`) so it can be toggled per environment from Chef.

When enabled, micromdm exposes the Basic-Auth `/mdm/migration-checkin`
endpoint used for NanoMDM migration dual-write: a dual-writer (mdmdirector)
mirrors Authenticate/TokenUpdate/CheckOut events from NanoMDM into this
MicroMDM's device store, keeping it fresh as a rollback target.


jessepeterson:
Thanks, but MicroMDM is in [maintenance mode](https://micromdm.io/blog/micromdm-v1-maintenance-mode/) and no longer accepting feature PRs.

PR #1128, opened 2026-09-12
fix(deps): bump golang.org/x/net past two real CVEs

boazgarty:
## Summary

`golang.org/x/net` is pinned at `v0.34.0` (a direct dependency), affected by:
- `GO-2026-5026` (ASCII-only Punycode-label rejection failure in `golang.org/x/net/idna`)
  — reachable via `micromdm.serve` → `httputil.ListenAndServe` and
  `DeclarativeManagementHTTPCaller` → `idna.ToASCII`
- `GO-2026-4918` (HTTP/2 transport infinite loop on a bad `SETTINGS_MAX_FRAME_SIZE`)
  — reachable via `DeclarativeManagementHTTPCaller`'s own `http.Client.Do` call chain

Both confirmed genuinely reachable via `govulncheck`'s call-graph analysis (not
just "pinned below the fixed version"), with real traces into this module's own
code. Bumped to `v0.55.0`, which covers both.

`golang.org/x/crypto` moves to `v0.51.0` as `x/net`'s own real transitive
requirement (confirmed via `go get golang.org/x/net@v0.55.0 && go mod tidy`) —
not because of any CVE claim against `x/crypto` itself.

## A real, disclosed CI caveat

The `go` directive move (`1.17` → `1.25.0`) is `x/net v0.55.0`'s own real
requirement. Go's `go.mod` parser only accepts the newer 3-component
`go 1.25.0` format starting at Go 1.21, and this repo's CI is currently
pinned to `go-version: ['1.20.x']` in `.github/workflows/CI.yml` — that
combination can't build at all (`invalid go version '1.25.0': must match
format 1.23`), independent of anything else in this diff. I don't have
permission to push a workflow-file change from this token, so I haven't
touched `CI.yml` here — **someone with write access will need to bump the
matrix's `go-version` to `1.25.x` (or later) for this PR's own CI to go
green.** Flagging it explicitly rather than leaving a silently-red check.

## Correction (edited after opening)

This PR originally also bumped `golang.org/x/crypto` directly, citing 9
further advisories ("7 CRITICAL"). **That claim was wrong and has been
removed.** All 9 are scoped to `golang.org/x/crypto/ssh`, `ssh/agent`, and
`ssh/knownhosts` — this module's real import graph never reaches any of
those (confirmed via `go list -deps ./...`: only `acme/autocert`,
`cryptobyte`, `pbkdf2`, and `pkcs12` are actually reachable). Sorry for the
noise on the first version of this diff.

## Test plan

- [x] `go build -v ./...` succeeds
- [x] `gofmt -s -l .` clean
- [x] `go test ./...` passes across the full suite, locally under Go 1.25.3
      (matches the module's own new floor; CI itself needs the version bump
      above before it can verify this on its own infrastructure)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jessepeterson:
Hello, thanks for your contribution! As the project README states:

> The [MicroMDM "v1" project is in maintenance mode](https://micromdm.io/blog/micromdm-v1-maintenance-mode/) and support officially ends at the end of 2025. You are encouraged to explore or migrate to [NanoMDM](https://github.com/micromdm/nanomdm) and other projects in the Nano-suite of projects.

The blog clarifies this is true of security issues as well.

Please read (or have your agent read) the organization's [contributing guide](https://github.com/micromdm/.github/blob/main/CONTRIBUTING.md) for future contributions — it appears it wasn't followed for this contribution.

Thanks again!

